PR #13, opened 2018-06-14
[134084](137265) move clr-pager into new repo

cat-mandres:


Wallabeng:
Please move the pager component into the new "clr-addons" folder, as this is the new root for our library

PR #22, opened 2018-06-18
New component added: paged-search-result-list

cat-mandres:
updates to clr-pager


src/clr-addons/clr-addons.module.ts
```diff
@@ -6,10 +6,11 @@
 
 import { NgModule } from '@angular/core';
 
-import { ClrViewEditSectionModule } from './view-edit-section/view-edit-section.module';
```
Wallabeng: Is there a reason for removing the filename?

src/clr-addons/paged-search-result-list/paged-search-result-list.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(ClrPagedSearchResultList);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
```
Wallabeng: Can we have more tests here please

src/clr-addons/clr-addons.module.ts
cat-mandres: intelliJ improvement suggestion

Wallabeng: just wondered because its not consistently removed everywhere e.g. one line below or in the src/dev files

src/clr-addons/paged-search-result-list/paged-search-result-list.spec.ts
cat-mandres: this component does not offer any methods - it is more or less just 

<pager>
<custom content>
<pager>

Do you have any specific tests in mind I could implement for this?

Wallabeng: Maybe a test where a page switch happens and then check whether content switched as expected.
Also please rename the test from "PagerComponent" to "ClrPagedSearchResultList", otherwise its confusing in the test results

PR #65, opened 2018-06-29
Add content panel documentation and demo

cat-mandres:


latest/src/app/documentation/demos/full-page-layouts/basepage-layout/basepage-layout.demo.html
```diff
@@ -2,18 +2,33 @@
...
-        </clr-button-group> 
+        </clr-button-group>
     </div>
-    <div class="content-container">
+    <div class="content-container" *ngIf="!withContentPanel">
         <div class="content-area">
             This is the page content
         </div>
     </div>
+    <clr-content-panel-container *ngIf="withContentPanel">
```
Wallabeng: This component should have class content-container attached by default.
And inside the component where the class "content-panel-outer-content" is attached, also the class "content-area" should be added.
This way the first child will have it's margin-top set to 0 by clarity default css.
Then you may also cleanup some css rules which overlap in those classes.
Sorry for that late finding, but just saw it in the demo.

Wallabeng:
You may also want to add the Content Panel page to the menu below the Sidebar Page Layout.

PR #98, opened 2018-07-25
Add card-active class for VU3 and mVAP

cat-mandres:
- fix bug when hovering cards

### DIFF
--- a/src/clr-addons/components.clr-addons.scss
+++ b/src/clr-addons/components.clr-addons.scss
@@ -25,3 +25,14 @@
     z-index: map-get($clr-layers, sidepanel) - 2;
   }
 }
+
+// Add class card-active
+.card {
+  &.card-active {
+    border-left: 0.25rem solid rgba(0, 148, 210, 1);
+
+    &.clickable:hover {
+      border-left: 0.25rem solid rgba(0, 148, 210, 1);
+    }
+  }
+}
--- a/src/clr-addons/components.clr-addons.scss
+++ b/src/clr-addons/components.clr-addons.scss
@@ -6,6 +6,9 @@
 @import './content-panel/content-panel';
 @import './breadcrumb/breadcrumb';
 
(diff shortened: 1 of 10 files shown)


src/clr-addons/components.clr-addons.scss
```diff
@@ -25,3 +25,14 @@
...
   }
 }
+
+// Add class card-active
+.card {
+  &.card-active {
+    border-left: 0.25rem solid rgba(0, 148, 210, 1);
+
+    &.clickable:hover {
+      border-left: 0.25rem solid rgba(0, 148, 210, 1);
```
Wallabeng: Please use/create variables here.
If you need new variables, create their default values here.
e.g. clr-card-active-border-color: rgba(0, 148, 210, 1) !default
The !default flag indicates, that it only updates the variable if it's not already present.
That way you can customize this variable in themes (vu3 and mvap) and it will not get overwritten here.
At the same time it provides default values for clarity light+dark

src/clr-addons/themes/mvap/css.overrides.scss
```diff
@@ -42,8 +42,19 @@ header,
...
 }
 
-.card.clickable:hover {
-  box-shadow: 0 0.125rem 0 0 $turquoise;
+.card {
+  &.card-active {
+    border-left-color: $turquoise;
+
+    &.clickable:hover {
+      border-left: 0.25rem solid $turquoise;
```
Wallabeng: card-active should be themed with variables, not overrides

src/clr-addons/themes/vu3/css.overrides.scss
```diff
@@ -39,6 +39,19 @@ header,
 
 .card {
   border-radius: $vu3-border-radius;
+
+  &.card-active {
+    border-left-color: $turquoise;
+
+    &.clickable:hover {
+      border-left: 0.25rem solid $turquoise;
```
Wallabeng: card-active should be themed with variables, not overrides

PR #122, opened 2018-08-02
Add substeps to flow component

cat-mandres:
- Adapt dev demos
- Create exclusive selectable card style, different from default clickable


src/clr-addons/flow-bar/flow-bar.ts
```diff
@@ -9,6 +9,8 @@ import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
 export interface ClrFlowBarStep {
   title: string;
   enabled: boolean;
+  activeSubStep?: ClrFlowBarStep;
```
Wallabeng: maybe only save the index of the activeSubStep, instead of the whole object, as every access to this variable (except one in html) needs to know the index of it.

PR #150, opened 2018-08-08
Implement clr-back-button

cat-mandres:


src/clr-addons/back-button/back-button.ts
```diff
@@ -0,0 +1,20 @@
...
+
+@Component({
+  selector: 'clr-back-button',
+  templateUrl: './back-button.html',
+})
+export class ClrBackButton {
+  constructor(private location: Location) {}
+
+  public back(): void {
+    this.location.back();
```
Wallabeng: I guess this will not work with iFrames. We need some logic with framebus.

PR #165, opened 2018-08-16
made radio button selection ring thinner

fachammer:
made radio button selection ring thinner to match VU3 mockups


fachammer:
still needs some other changes

PR #186, opened 2018-09-04
add progress spinner directive

mbrunnbauer:


src/clr-addons/progress-spinner/progress-spinner.scss
```diff
@@ -0,0 +1,16 @@
...
+  background-color: rgba(255, 255, 255, 0.8);
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  z-index: 2;
```
cat-mandres: I think this should be higher. if we use this inline spinner inside a modal for example the overlay would not be shown above the modal. I dont think any other element should be higher than the loading spinner so we can set an absurdly high value here, like 99999?

src/clr-addons/progress-spinner/progress-spinner.scss
```diff
@@ -0,0 +1,16 @@
+.loading-overlay {
+  background-color: rgba(255, 255, 255, 0.8);
```
cat-mandres: should use the $clr-white color variable if thats possible to be combined with the opacity

src/clr-addons/progress-spinner/progress-spinner.scss
```diff
@@ -0,0 +1,16 @@
...
+  width: 100%;
+  height: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  z-index: 2;
+  display: flex;
+}
+
+.loading-spinner {
```
cat-mandres: remove empty?

src/clr-addons/progress-spinner/index.ts
```diff
@@ -0,0 +1,6 @@
+/*
+ * Copyright (c) 2018 Porsche Informatik. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+export * from './progress-spinner.module';
```
Wallabeng: I'm not sure if we also need to export the directive file here, as we never use the class itself directly.
For now let's try without it and see what happens in the web demo.

PR #190, opened 2018-09-05
rename loadingState attribute of progress spinner

ghost:


src/clr-addons/progress-spinner/progress-spinner.directive.ts
```diff
@@ -10,7 +10,7 @@ import { Directive, ElementRef, Input, OnChanges, OnInit, Renderer2, SimpleChang
 })
 export class ClrProgressSpinnerDirective implements OnInit, OnChanges {
   static readonly MINIMUM_VISIBLE_DURATION = 200;
-  @Input() loadingState: boolean;
+  @Input() clrProgressSpinner: boolean;
```
Wallabeng: Please name the variable "showSpinner", but rename it to the outer world via Input('clrProgressSpinner').
That way it's more logical inside the component.

PR #192, opened 2018-09-05
added a fix for hovering over a selectable card in firefox

RubenGantner:


src/clr-addons/components.clr-addons.scss
```diff
@@ -78,6 +78,27 @@ body,
     @include card-active-after();
   }
 
+  // firefox needs this, because the hover in @media (hover: hover) is not supported
+  @-moz-document url-prefix() {
```
Wallabeng: Please put the content of both queries in a mixin and use it in both to prevent multiple maintenance

PR #218, opened 2018-09-20
adapt treetable

mbrunnbauer:


src/clr-addons/treetable/treetable-row.ts
```diff
@@ -4,43 +4,51 @@
...
-
-  ngAfterContentInit(): void {
-    this.initIsExpandable();
-    this.ttRows.changes.subscribe(() => this.initIsExpandable());
-  }
-
-  private initIsExpandable(): void {
-    this.isExpandable = this.ttRows.length > 1;
-  }
+  @Input('clrExpandable') expandable = false;
```
Wallabeng: do we have a usecase where the row has children, but should not be expandable?
Because this way you have to specify expandable on every row.

If we have such a use case, I bet this is the rarer case, so the default value should be true and should only be set to false for those exceptional cases.

PR #247, opened 2018-11-21
Upgrade to 1.0.0-beta.2

Wallabeng:


src/clr-addons/treetable/treetable.scss
```diff
@@ -1,21 +1,235 @@
...
+          }
+
+          .treetable-cell:first-of-type {
+            padding-left: $base-cell-indentation;
+          }
+          .treetable-expandable-caret {
+            left: $base-caret-indentation;
+          }
+
+          clr-tt-row {
```
Wallabeng: replace with class

PR #273, opened 2018-12-05
Numeric field - create custom validators and fix a bug with the unit …

cat-mandres:
…symbol.


src/clr-addons/numericfield/numeric-field.ts
```diff
@@ -191,15 +195,15 @@ export class ClrNumericField implements OnInit {
       this.renderer.setStyle(inputWrapper, 'width', inputWidth + 'px');
```
Wallabeng: Just to keep in mind: with this whole fixed width stuff, we can't use responsive inputs. Which for now is not a problem, as clarity doesn't support it either out of the box.

PR #300, opened 2018-12-18
Treetable: Set max width of first cell of each row.

cat-mandres:


src/clr-addons/treetable/renderer/main-renderer.ts
```diff
@@ -22,6 +22,10 @@ export class TreetableMainRenderer<T = any> implements AfterViewInit {
 
   ngAfterViewInit(): void {
```
Wallabeng: this needs to be ngAfterViewChecked, as the rows could change during component lifetime, so new rows also get the classes

PR #327, opened 2019-01-08
update documentation to v4

fschickermueller:


latest/src/app/news/news.component.html
```diff
@@ -34,5 +34,14 @@ <h1>{{releaseNumber}} Changelog</h1>
                 </li>
             </ul>
         </section>
+        <section class="nav-group collapsible">
```
Wallabeng: please move this to the top of list, so the version in the menu is descending

PR #332, opened 2019-01-24
Support ChangeDetectionStrategy.OnPush for Progress Spinner Component

cat-mandres:


cat-mandres:
reviewed by marcel brunnbauer

PR #344, opened 2019-02-06
contentPanel : increase timeout for window resize

ziedom12:
- increase timeout for window resize event when closing the panel

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clr-addons",
-  "version": "5.0.1",
+  "version": "5.0.2",
   "private": true,
   "scripts": {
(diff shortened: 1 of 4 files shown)


src/clr-addons/content-panel/content-panel.ts
```diff
@@ -53,10 +53,10 @@ export class ClrContentPanel implements OnInit {
     this._open = false;
     this.closed.emit();
     // this event is used by the tree table to adjust the colomns width
-    // 220 is for the animation to finish.
+    // 200 is for the animation to finish + 60 safety marge.
     setTimeout(() => {
       window.dispatchEvent(new Event('resize'));
-    }, 220);
+    }, 260);
```
Wallabeng: Can we use Animation callbacks for these? As timeouts will work on some clients, but on slower clients won't.
https://angular.io/guide/transition-and-triggers#animation-callbacks

ziedom12: sounds a good idea, I will try it

ziedom12:
Hi, Should I change the version as well ?

Wallabeng:
> Hi, Should I change the version as well ?

yes you can, no additional changes are planned this week

src/clr-addons/content-panel/content-panel.ts
ziedom12: is 5.0.1-alpha-1 fine ?

ziedom12:
5.0.1-alpha-1 fine ?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clr-addons",
-  "version": "5.0.1",
+  "version": "5.0.1-alpha-1",
```
Wallabeng: 5.0.2 please
we ditched the concept of alphas and betas, as the project is too small for that

PR #350, opened 2019-02-18
Set max width for view edit section, adapt forms css overrides, relea…

cat-mandres:
…se version 5.1.2

### DIFF
--- a/src/clr-addons/components.clr-addons.scss
+++ b/src/clr-addons/components.clr-addons.scss
@@ -278,7 +278,7 @@ clr-date-container {
   }
 }
 
-// Set the width of all inputs except time & not inside a .clr-input-group to a 100%
+// Set the width of all inputs except time to 100%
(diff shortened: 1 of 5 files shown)


src/clr-addons/components.clr-addons.scss
```diff
@@ -249,20 +249,51 @@ clr-checkbox-container[readonly] .clr-checkbox-wrapper input[type='checkbox'] {
...
+  max-width: 100%;
+
+  // Set all input widths inside a clr-input-group to 100% minus the icons width
+  > .clr-input {
+    width: calc(100% - 1.5rem);
   }
 }
 
-.clr-control-container .clr-textarea-wrapper textarea {
+// Set the width of all inputs except time & not inside a .clr-input-group to a 100%
```
Wallabeng: "& not inside a .clr-input-group"
seems to be a leftover from development, right?

PR #353, opened 2019-02-18
 Update numeric-field.ts

proggler23:
implement rounding + fix several issues

### DIFF
--- a/src/clr-addons/numericfield/numeric-field.ts
+++ b/src/clr-addons/numericfield/numeric-field.ts
@@ -40,6 +40,7 @@ export class ClrNumericField implements OnInit, OnDestroy, AfterViewChecked {
   @Input('clrUnitPosition') unitPosition: string = 'right';
   @Output('clrNumericValueChange') numericValueChanged = new EventEmitter<number>();
 
+  private displayValue: string = '';
   private originalValue: number = NaN;
   private _numericValue: number;
   private inputChangeListener: () => void;
@@ -149,10 +150,8 @@ export class ClrNumericField implements OnInit, OnDestroy, AfterViewChecked {
     // Sometimes the value changes because we cut off decimal places
     setTimeout(() => {
       this.updateInput(
-        this.formatNumber(this._numericValue.toString().replace(new RegExp('[.]', 'g'),
-          this.decimalSeparator),
-          true),
-        false);
+        this.formatNumber(this._numericValue.toString().replace(new RegExp('[.]', 'g'), this.decimalSeparator), true)
+        , false);
     }, 1);
   }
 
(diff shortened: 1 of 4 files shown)


src/clr-addons/numericfield/numeric-field.ts
```diff
@@ -120,11 +122,14 @@ export class ClrNumericField implements OnInit, OnDestroy, AfterViewChecked {
           if (event.keyCode === BACK_KEYCODE && value.substr(cursorStart - 1, 1) === this.groupingSeparator) {
             event.target.value = value.substring(0, cursorStart - 2) + value.substring(cursorStart - 1, value.length);
             event.target.selectionStart = event.target.selectionEnd = cursorStart - 1;
+            return false;
```
Wallabeng: is this really needed? i want to prevent aborting the normal event flow where possible.

src/clr-addons/numericfield/numeric-field.ts
```diff
@@ -120,11 +122,14 @@ export class ClrNumericField implements OnInit, OnDestroy, AfterViewChecked {
...
           } else if (event.keyCode === DEL_KEYCODE && value.substr(cursorStart, 1) === this.groupingSeparator) {
             event.target.value = value.substring(0, cursorStart + 1) + value.substring(cursorStart + 2, value.length);
             event.target.selectionStart = event.target.selectionEnd = cursorStart + 1;
+            return false;
```
Wallabeng: same here

src/clr-addons/numericfield/numeric-field.ts
```diff
@@ -120,11 +122,14 @@ export class ClrNumericField implements OnInit, OnDestroy, AfterViewChecked {
...
           }
         }
+        this.originalValue = NaN;
```
Wallabeng: I guess you did the return false, because of this line. But is this line needed?
The value will get emitted anyways as "userEntered" will be true.

src/clr-addons/numericfield/numeric-field.ts
proggler23: needed for an issue, that when you are on the left to a grouping-seperator and press DEL, 2 digits are removed

src/clr-addons/numericfield/numeric-field.ts
proggler23: needed for an issue, that when you are on the left to a grouping-seperator and press DEL, 2 digits are removed

src/clr-addons/numericfield/numeric-field.ts
proggler23: might be unnecessary, but this also more clearly indicates a 'dirty' input field

PR #363, opened 2019-02-26
[162954](162955) Fix numeric field reset on entering 0

proggler23:


src/clr-addons/numericfield/numeric-field.ts
```diff
@@ -222,9 +222,11 @@ export class ClrNumericField implements OnInit, OnDestroy, AfterViewChecked {
           }
           indexDecimalSep = j;
         }
-        if (char === '0' && removeLeadingZeros && ((result.length === 0 && (j + 1) !== value.length) || result === NEGATIVE)) {
-          ignoredChars++;
-          continue;
+        if (char === '0' && removeLeadingZeros) {
```
Wallabeng: Code is ok.
But please add some documentation as it needs some time to understand.

PR #381, opened 2019-03-15
[160153](164606) fix notifications to wrap the line also on page breaks

moser96:
detail: if the message keys are specified in the following way:
"message first line \n message second line" the notification component should wrap correctly


ziedom12:
you can change the version number within the same commit also.

PR #410, opened 2019-04-17
QuickList component

ziedom12:


src/clr-addons/quick-list/quick-list.component.html
```diff
@@ -0,0 +1,15 @@
+<div class="clr-form-control">
+  <div class="clr-row">
+    <div class="clr-col-12">
+      <ng-container *ngFor="let value of values; index as i; trackBy: trackById">
+        <clr-add-option [allValues]="allValues" [excludedValues]="values" (remove)="onRemoveValue(i)"
+                        [value]="value" (valueChanged)="onValueChanged($event, i)"
+                        [blankOption]="blankOption">
+        </clr-add-option>
+      </ng-container>
+      <div class="clr-row pt-0-5 pb-0-5">
```
Wallabeng: pb should not be used, as the next element needs to have a padding top if needed

src/clr-addons/quick-list/add-option/add-option.component.html
```diff
@@ -0,0 +1,17 @@
...
+      <select clrSelect name="options" [ngModel]="selectedValue" (ngModelChange)="selectedValue = $event">
+        <option *ngFor="let val of getValues()"
+                [selected]="val.id === selectedValue" [disabled]="!val.value" [ngValue]="val.id">
+          {{val.label}}
+        </option>
+      </select>
+    </clr-select-container>
+  </div>
+  <div class="clr-col-2 pl-0">
+    <button class="btn btn-icon btn-link btn-thrash">
```
Wallabeng: Trash button should be disabled on last visible item when this quick list is mandatory.
For this a new input is needed to give users the possibility to flag this component as mandatory.

Wallabeng: trash button doesn't work at all...

src/clr-addons/quick-list/add-option/add-option.component.ts
```diff
@@ -0,0 +1,38 @@
...
+    this.valueChanged.emit(this.value);
+  }
+
+  get selectedValue(): string {
+    return this.value.id;
+  }
+
+  getValues() {
+    const excludedIds = this.excludedValues.filter(val => val.id !== this.value.id).map(val => val.id);
+    return [this.blankOption, ...this.allValues].filter(val => !excludedIds.includes(val.id));
```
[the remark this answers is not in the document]
ziedom12: If the blank option label is the widest amongst all the other labels, the select box would be changing size when adding and changing selected options. I would rather have it always. an option showing and hiding effects to select component size

src/clr-addons/quick-list/add-option/add-option.component.html
```diff
@@ -0,0 +1,17 @@
...
+    <clr-select-container>
+      <select clrSelect name="options" [ngModel]="selectedValue" (ngModelChange)="selectedValue = $event">
+        <option *ngFor="let val of getValues()"
+                [selected]="val.id === selectedValue" [disabled]="!val.value" [ngValue]="val.id">
+          {{val.label}}
+        </option>
+      </select>
+    </clr-select-container>
+  </div>
+  <div class="clr-col-2 pl-0">
```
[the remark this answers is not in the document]
ziedom12: removing this col-2 breaks the layout in some cases.

src/clr-addons/quick-list/add-option/add-option.service.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2018 Porsche Informatik. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export interface ClrQuickListValue<T> {
```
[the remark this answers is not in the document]
ziedom12: well, we are checking the selected option with id. this is useful in case of redundant values. Ids are more sure to be unique, isn't it ?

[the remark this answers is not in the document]
Wallabeng: ok then id and label should be enough here and you don't need the value property right?
as the id should be emitted also, as it's more accurate

src/clr-addons/quick-list/add-option/add-option.component.ts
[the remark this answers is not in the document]
Wallabeng: the size of the quick list should always be 100% of the width it was given (not dependent on its content)
e.g. you set col-3 in the demo, the selects should cover the col-3 space
and if you set col-12 they should be as wide as col-12 is in that location.
as our standard selects also work that way and have 100% already set to them.

src/clr-addons/quick-list/add-option/add-option.component.html
Wallabeng: click handler needs to be set on the button, not the icon, otherwise this doesn't work in firefox

src/clr-addons/quick-list/quick-list.component.scss
```diff
@@ -0,0 +1,11 @@
+.pl-0-75 {
+  padding-left: 0.75rem;
+}
+
+.pb-0-5 {
+  padding-bottom: 0.5rem;
+}
+
+.pt-0-5 {
```
Wallabeng: please don't name our classes in the naming scheme of bootstrap to avoid future collisions.

src/dev/src/app/quick-list/quick-list.demo.ts
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (c) 2018 Porsche Informatik. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Component, OnInit } from '@angular/core';
+import { ClrQuickListValue } from '../../../../clr-addons/quick-list/add-option/add-option.service';
```
Wallabeng: this should be imported from the package @porscheinformatik/clr-addons

src/dev/src/app/quick-list/quick-list.demo.html
```diff
@@ -0,0 +1,31 @@
+<h2>Quick List</h2>
+<div class="clr-example">
+  <div class="main-container">
+    <div class="clr-row">
+      <div class="clr-col-1"></div>
```
Wallabeng: why is this div here

src/dev/src/app/quick-list/quick-list.demo.html
```diff
@@ -0,0 +1,31 @@
...
+      <clr-quick-list class="clr-col-3"[allValues]="possibleActivityResults" [(values)]="selectedActivityResults"
```
Wallabeng: this clr-col-3 has no effect right now, but it should.
the selects should be as wide as this clr-col class describes

src/dev/src/app/quick-list/quick-list.demo.html
Wallabeng: better use it in a form example

src/dev/src/app/quick-list/quick-list.demo.html
ziedom12: I failed in making the select component cover all the reserved width, any hint please ?

src/dev/src/app/quick-list/quick-list.demo.html
ziedom12: ok will give it a try, thanks

src/clr-addons/quick-list/add-option/add-option.component.scss
```diff
@@ -0,0 +1,15 @@
+.btn-thrash {
+  margin-bottom: 0;
+  margin-top: 0.75rem;
+}
+.quick-list-input {
+  flex: 1;
+}
+
+.quick-list-trash {
```
Wallabeng: not used

src/clr-addons/quick-list/quick-list.component.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Component, EventEmitter, Input, Output } from '@angular/core';
+import { CLR_BLANK_OPTION, ClrQuickListValue } from './add-option/add-option.service';
+
+@Component({
+  selector: 'clr-quick-list',
+  templateUrl: './quick-list.component.html',
+  styleUrls: ['./quick-list.component.scss'],
+})
+export class ClrQuickListComponent<T> {
+  @Input() blankOption: ClrQuickListValue<T> = CLR_BLANK_OPTION;
```
Wallabeng: all public input and output parameters need to be prefixed with clr according to our styleguide. Only needed in this class, as the option component is not public
e.g. @Input('clrBlankOption') blankOption

src/clr-addons/quick-list/quick-list.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  @Input() allValues: Array<ClrQuickListValue<T>> = [this.blankOption];
+  @Input() values: Array<ClrQuickListValue<T>> = [this.blankOption];
+  @Input() addLabel: string = 'ADD';
+  @Input() mandatory: boolean = false;
+  @Output() valuesChanged = new EventEmitter<Array<ClrQuickListValue<T>>>();
+  @Output() addClicked = new EventEmitter<void>();
```
Wallabeng: this event emitter is not needed, as valuesChanged already emits every change on the values

src/clr-addons/quick-list/add-option/add-option.component.html
```diff
@@ -0,0 +1,19 @@
+<div class="quick-list-input">
+  <clr-select-container>
+    <select class="w-100" clrSelect name="options" [(ngModel)]="selectedValue" (ngModelChange)="selectedValue = $event">
```
cat-mandres: get rid of the (ngModelChange) - you already have 2 way binding there.

why is there a w-100? selects should have 100% width automatically.

src/clr-addons/quick-list/add-option/add-option.component.scss
```diff
@@ -0,0 +1,11 @@
+.btn-thrash {
+  margin-bottom: 0;
+  margin-top: 0.75rem;
+}
+.quick-list-input {
+  flex: 1;
+}
+
+:host(.quick-list-option) {
```
cat-mandres: host selector in css shouldnt be used anymore - use the host property your @Component annotation to apply certain classes to your host element

cat-mandres: or @HostBinding annotation.
you should then be able to get rid of the :host()

src/clr-addons/quick-list/add-option/add-option.service.ts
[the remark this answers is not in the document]
cat-mandres: still open?

src/clr-addons/quick-list/add-option/add-option.component.html
ziedom12: removing the w-100 makes the select panel only cover the largest available label

cat-mandres: I see yes you are right, our custom CSS which sets selects to 100% width only works inside forms, you are not using forms inside this component

src/clr-addons/quick-list/add-option/add-option.component.scss
ziedom12: is this a proper solutions now ? 
@HostBinding('style.display') display = 'flex';

src/clr-addons/components.clr-addons.scss
```diff
@@ -16,6 +16,7 @@
 @import './combobox/combobox.clarity';
 @import './date-time-container/date-time-container';
 @import './searchfield/search-field';
+@import './quick-list/add-option/add-option';
```
cat-mandres: you have 2 new SCSS files but only import one here

src/clr-addons/quick-list/quick-list.component.ts
ziedom12: done

src/clr-addons/quick-list/quick-list.component.ts
ziedom12: could be beneficial in case of automatic scoll down when clicking add

src/dev/src/app/quick-list/quick-list.demo.html
ziedom12: div deleted

src/clr-addons/quick-list/add-option/add-option.component.html
ziedom12: so, I leave it now

src/clr-addons/quick-list/add-option/add-option.component.scss
ziedom12: fixed

src/clr-addons/components.clr-addons.scss
ziedom12: we only use one now

src/clr-addons/quick-list/add-option/add-option.service.ts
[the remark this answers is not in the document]
Wallabeng: yes this is still open, there was no answer or solution. so this shouldn't be closed

src/clr-addons/quick-list/add-option.service.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2018 Porsche Informatik. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export interface ClrQuickListValue<T> {
+  id: string;
+  label?: string;
+  value: T;
```
Wallabeng: question remains: is this value needed? as it's not needed in the whole component

src/dev/src/app/quick-list/quick-list.demo.html
```diff
@@ -0,0 +1,30 @@
+<h2>Quick List</h2>
+<div class="clr-example">
+  <div class="main-container">
+    <div class="clr-row">
+      <clr-quick-list class="clr-col-6 scroll" [clrAllValues]="possibleActivityResults" [(clrValues)]="selectedActivityResults"
```
Wallabeng: As requested, please add a real life example inside a form with at least on normal input field with label/value. I don't think this will work that easy, right now.
I guess this will be used alongside other form elements. So I want to check if it fits in this layout.

src/clr-addons/quick-list/add-option.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+import { Component, EventEmitter, Input, Output } from '@angular/core';
+import { ClrQuickListValue } from './add-option.service';
+
+@Component({
+  selector: 'clr-add-option',
+  templateUrl: './add-option.html',
+  host: { '[class.quick-list-option]': 'true' },
+})
+export class ClrAddOption<T> {
```
Wallabeng: We would also need some unit tests for those two components please.

src/clr-addons/quick-list/add-option.service.ts
ziedom12: This value not needed within the component. I assume it's there to keep tracking of the selected values, the real object behind the the labels themselves.

Wallabeng: could be done with id also and would be more accurate right? And why do you assume? You wrote the component, so there must be a reason behind it.

ziedom12: I think I miss explained my self, so again : 
- we use the ids to track the selected options.
- the values contain the real data behind the ids, we don't use them in the component. but the developer will use them directly, he will not have to check again for the selected ids, and look for the corresponding objects.

src/clr-addons/quick-list/add-option.ts
ziedom12: I added unit tests only for QuickList component

PR #491, opened 2019-06-27
fix quick-list - empty list handling when mandatory

adichioran:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clr-addons",
-  "version": "6.0.1",
+  "version": "6.0.2",
```
Wallabeng: please remove the version, as i also have a change incoming and will generate one version out of it then

PR #531, opened 2019-08-21
focus select on item added + keep focus on change

adichioran:


src/clr-addons/quick-list/quick-list.ts
```diff
@@ -4,34 +4,56 @@
...
   templateUrl: './quick-list.html',
 })
-export class ClrQuickList<T> implements OnInit {
+export class ClrQuickList<T> implements OnInit, AfterViewChecked {
   @Input('clrBlankOption') blankOption: ClrQuickListValue<T> = CLR_BLANK_OPTION;
   @Input('clrAllValues') allValues: Array<ClrQuickListValue<T>> = [this.blankOption];
   @Input('clrMandatory') mandatory: boolean = false;
   @Input('clrValues') values: Array<ClrQuickListValue<T>> = [];
   @Input('clrAddLabel') addLabel: string = 'ADD';
-  @Input('clrControlClasses') controlClasses = 'clr-col-md-10';
```
Wallabeng: i would not recommend removing this, as it's a breaking change and the label also provides md-2 as default, so the layout should be fine without developer-interaction.

src/clr-addons/quick-list/quick-list.ts
```diff
@@ -4,34 +4,56 @@
...
+  focusId: string = null;
+
   ngOnInit(): void {
     if (this.values.length === 0 && this.mandatory) {
       this.values.push(this.blankOption);
     }
   }
 
+  ngAfterViewChecked(): void {
+    if (!!this.focusId) {
```
Wallabeng: this whole focusing can be moved to the option itself, as the "set selectedValue" is triggered on change.
that way you don't need to remember any id, but just focus in the method "set selectedValue" (maybe with a setTimeout to wait for everything to finish)

src/clr-addons/quick-list/quick-list.ts
adichioran: if i provide this as a default, the field will not be usable in vertical forms as this classes add a padding to the field

src/clr-addons/quick-list/quick-list.ts
adichioran: i need to set the focus on 2 events
add Option
value changed 

the problem is for add option which is an event triggered from outside the option

cat-mandres: is set selectedValue not triggered ONCE after you click on add option? I would think you can just focus in the set selectedValue method of the option itself as marcel described. if not please describe in more detail why that doesnt work.
(if the setter is actually not called for whatever reason, you can still focus on ngAfterViewInit? )

adichioran: the add option button adds a new selection field
that selection field needs to be focused automatically
the option itsself cannot know if its the last option in the list of options in order to set the focus on itsself

adichioran: i tried many solutions, including this one, but they were not complete, or did not work at all

src/clr-addons/quick-list/quick-list.ts
```diff
@@ -22,6 +32,8 @@ export class ClrQuickList<T> implements OnInit {
   @Output('clrValuesChanged') valuesChanged = new EventEmitter<Array<ClrQuickListValue<T>>>();
   @Output('clrEmptyOptionAdded') emptyOptionAdded = new EventEmitter<void>();
 
+  @ViewChildren(ClrAddOption) options: QueryList<ClrAddOption<T>>;
```
cat-mandres: unused?

remove unused imports aswell

PR #558, opened 2019-10-14
Fixed notification dialog width for SM display size

gurumaxi:


src/clr-addons/notification/notification.scss
```diff
@@ -1,3 +1,5 @@
+@import '../../../node_modules/@clr/ui/src/layout/grid/variables/variables.clarity';
```
cat-mandres: you (most likely, please test) dont have to import this to access clr variables in this repo - we import all of them automatically and use/override them.

src/clr-addons/notification/notification.scss
```diff
@@ -78,3 +80,10 @@
     }
   }
 }
+
+@media screen and (max-width: map-get($clr-grid-breakpoints, sm)) {
```
cat-mandres: why dont add the media query inside the already existing selector for the wrapper? see line 18

src/clr-addons/notification/notification.scss
```diff
@@ -78,3 +80,10 @@
...
+  .notification__wrapper {
+    width: calc(100% - 2rem) !important;
```
Wallabeng: if you place this width definition after the original with 400px, this should overwrite it without having to use !important (as this is bad practice)

PR #564, opened 2019-10-30
Adjusted content-header padding and page container padding for XS display size

gurumaxi:


src/clr-addons/components.clr-addons.scss
```diff
@@ -52,6 +60,20 @@ body,
...
+    }
+  }
+
+  @media screen and (max-width: map-get($clr-grid-breakpoints, sm)) {
+    display: block;
+    overflow: hidden;
+    padding: 0 0.5rem;
+
+    > * {
+      float: left;
```
Wallabeng: why is this needed? if you do this, no item can be right aligned in the header e.g. command bar (although there wouldn't be much alignment on such small screens, but we force everyone to left alignment)
just remove the "display: block" and "overflow:hidden" from content-header and the h2 should already reduce it's size so no overflow should happen on the content-header right?

src/clr-addons/components.clr-addons.scss
```diff
@@ -347,3 +370,11 @@ input:not([type='time']) {
   text-overflow: ellipsis;
   white-space: nowrap;
 }
+
+@media screen and (max-width: map-get($clr-grid-breakpoints, sm)) {
+  .main-container .content-container .content-area,
+  .content-area-footer,
```
Wallabeng: for .content-area-footer this will actually increases top/bottom margin (from .25rem). is this intended?

src/clr-addons/components.clr-addons.scss
gurumaxi: Buttons and command bars are not intended to be right aligned for screens widths < 576px. They need to be wrapped in a second line.

src/clr-addons/components.clr-addons.scss
gurumaxi: This is intended, as the padding is needed for the users on smartphones to feel more comfortable.

src/clr-addons/components.clr-addons.scss
Wallabeng: Just talked with Dieter Schrottshammer and it's not intended to wrap the header at all

PR #572, opened 2019-11-25
Change layout to support scrolling & sticky footer on smartphones

cat-mandres:
Especially fix issues which only occur on safari

### DIFF
--- a/src/clr-addons/components.clr-addons.scss
+++ b/src/clr-addons/components.clr-addons.scss
@@ -37,14 +37,6 @@ body,
   }
 }
 
-.main-container {
-  @media screen and (max-width: map-get($clr-grid-breakpoints, sm)) {
-    padding-top: $clr-header-height;
-    display: block;
-    overflow-y: auto;
-  }
-}
-
 .content-header {
   padding: 0.5rem 0.5rem 0 1rem;
   display: flex;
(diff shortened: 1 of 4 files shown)


src/clr-addons/components.clr-addons.scss
```diff
@@ -367,4 +380,25 @@ input:not([type='time']) {
   .content-panel-content-wrapper {
     padding: 0.5rem;
   }
+
+  //prevents the upper body part from being hidden by the fixed header
+  body {
+    padding-top: $clr-header-height;
```
Wallabeng: padding on the body, will move the whole page downwards with empty space at the top (at least in my applications)

src/clr-addons/components.clr-addons.scss
```diff
@@ -367,4 +380,25 @@ input:not([type='time']) {
...
+  }
+
+  //main-container needs the display:block property in order to work properly on safari
+  .main-container {
+    display: block;
```
Wallabeng: with those changes on the layout, u-main-container stops working correctly.
plus it's a heavy change of the standard clarity layout which I don't really want to do!
if there are any major issues on specific browsers we need to fix it in clarity itself.
just tested systemmanagement in safari and iphone 7 and it works perfectly fine with current implementation on DEV

cat-mandres: As stated in a VU3MY userstory, the title-bar needs to be scrollable, whereas header and footer need to be sticky. Due to the current page layout, this can only be achieved by removing the display:flex property and fixing header and footer.

Wallabeng: hmm ok, can we then please create a separate file called mobile.scss and put all the stuff inside there to find it easier.
additionally the media selector only has to be defined at the first line

PR #574, opened 2019-11-27
Change card top margin on extra small devices

cat-mandres:


src/clr-addons/components.clr-addons.scss
```diff
@@ -168,6 +168,10 @@ body,
     border-top-style: dashed;
     border-right-style: dashed;
   }
+
+  @media screen and (max-width: map-get($clr-grid-breakpoints, sm)) {
```
Wallabeng: please move to mobile.scss

PR #577, opened 2019-12-05
Adapt combobox to display default select on small width devices

cat-mandres:


src/clr-addons/combobox/combobox.ts
```diff
@@ -251,4 +254,27 @@ export class ClrCombobox<T> implements OnInit, AfterContentInit, OnDestroy {
...
+      case MobileBehaviourMode.FORCE_SELECT:
+        return true;
+      case MobileBehaviourMode.FORCE_AUTOCOMPLETE:
+        return false;
+      default:
+        return !this.allowUserEntry;
+    }
+  }
+
+  getOptionsAsArray(): Array<ClrOption<T>> {
```
Wallabeng: pls test (if not done already) if this also works options which change during input (backend filtering).
deferred initial loading is already tested, as I can see in demo code, so this should basically work

cat-mandres: tested --> works as expected

& build also fixed with next push

PR #579, opened 2019-12-09
Adapt signpost display change for display on smartphone

cat-mandres:
+ the !importants are necessary because of the dynamic positioning of the signpost

### DIFF
--- a/src/clr-addons/mobile.scss
+++ b/src/clr-addons/mobile.scss
@@ -70,4 +70,32 @@
   .card {
     margin-top: 0.5rem;
   }
+
+  .signpost {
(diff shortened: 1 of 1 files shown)


src/clr-addons/mobile.scss
```diff
@@ -70,4 +70,32 @@
   .card {
     margin-top: 0.5rem;
   }
+
+  .signpost {
+    .signpost-content {
```
Wallabeng: tested it with signpost position right-middle and it changed the position of the signpost.
why is all this stuff even needed? visually it only changes the width, for which the 'width css' is enough

PR #609, opened 2019-12-12
Exported enum MobileBehaviourMode to make it available from the outside

cat-mandres:


src/dev/src/app/combobox/combobox.demo.ts
```diff
@@ -6,7 +6,7 @@
 import { Component } from '@angular/core';
 import { of } from 'rxjs';
 import { delay } from 'rxjs/operators';
-import { MobileBehaviourMode } from '../../../../clr-addons/combobox/utils/constants';
+import { MobileBehaviourMode } from '../../../../clr-addons';
```
Wallabeng: should be "from '@clr/addons'", right?

PR #683, opened 2020-02-11
Fix multiple decorators for ivy compiler

furti:
In Ivy, it is not allowed to have a @Input and @ContentChild decorator on
the same property. We should use two dedicated properties and check which
one is set when using them.

https://github.com/angular/angular/issues/31398

### DIFF
--- a/src/clr-addons/paged-search-result-list/paged-search-result-list.ts
+++ b/src/clr-addons/paged-search-result-list/paged-search-result-list.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2019 Porsche Informatik. All Rights Reserved.
+ * Copyright (c) 2018-2020 Porsche Informatik. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -35,8 +35,10 @@ export class ClrPagedSearchResultList {
    * The template how each item should be displayed.
    */
   @Input('clrItemTemplate')
+  itemTemplateInput: TemplateRef<any>;
+
   @ContentChild(TemplateRef, { static: true })
-  itemTemplate: TemplateRef<any>;
+  itemTemplateContent: TemplateRef<any>;
 
   /**
    * The position of the pager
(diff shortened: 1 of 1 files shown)


jerone:
> Ivy and latest angular is not yet supported by clarity, so we can’t neither.

Support for Ivy and latest Angular was include in _pre-release_ [v3.0.0-next.8](https://github.com/vmware/clarity/releases/tag/3.0.0-next.8), released 3 days ago. Maybe this is not the correct issue to ask, but what will be the timespan for Clarity Addons to upgrade?

furti:
@Wallabeng when compiling with Ivy this was one of the problems that could easily be fixed without breaking compatibility with older versions. So I created a pull request for it. The tests pass so I think it does not break anything. But I have not tested it, as we are not using this component.

But maybe the @Input decorator can be removed completely? It is not even documented that it can be used this way. This would break compatibility with older versions.

Wallabeng:
> > Ivy and latest angular is not yet supported by clarity, so we can’t neither.
> 
> Support for Ivy and latest Angular was include in _pre-release_ [v3.0.0-next.8](https://github.com/vmware/clarity/releases/tag/3.0.0-next.8), released 3 days ago. Maybe this is not the correct issue to ask, but what will be the timespan for Clarity Addons to upgrade?

Yes clarity currently supports it in a beta release, but we will wait for a finally released version.
From there on we will plan the upgrade, but there is no timeline yet, as we can‘t predict the final release date for clarity.

> @Wallabeng when compiling with Ivy this was one of the problems that could easily be fixed without breaking compatibility with older versions. So I created a pull request for it. The tests pass so I think it does not break anything. But I have not tested it, as we are not using this component.
> But maybe the @input decorator can be removed completely? It is not even documented that it can be used this way. This would break compatibility with older versions.

If you just upgrade to latest angular and activate Ivy, the outcome of clarity and clarity addons is not predictable, as with a quick test they are not compatible with Ivy. Thats why I want to wait for a final release of clarity before changing how things work, as the behaviour could be optimized.
Currently I wouldn‘t recommend upgrading to Ivy at all, with released versions.

Wallabeng:
Closed in favor of #705

PR #705, opened 2020-03-11
Upgrade to Angular 9 and Clarity 3

cat-mandres:
+ Replace all SASS variables with CSS properties
+ Adapted REM values in some places because the base rem values was changed to 20px from 24px
+ Fixed other minor styling issues
+ Mark Combobox as Deprecated
+ Implement ControlValueAccessor for ClrNumericField to support reactive forms
+ Add Clr Data List Validator that only accepts values from a predefined array of strings
+ Increase to Version 7


src/clr-addons/components.clr-addons.scss
```diff
@@ -297,13 +303,13 @@ form,
   }
 }
 
-// Date picker should never be bigger than 6rem including the icon
+// Date picker should never be bigger than 7.2rem including the icon
```
Wallabeng: please just remove the value here and reformulate the comment

src/clr-addons/combobox/combobox.ts
```diff
@@ -89,7 +95,7 @@ export class ClrCombobox<T> implements OnInit, AfterContentInit, OnDestroy {
   selectedValue: T = this.preselectedValue;
 
   constructor(
-    private ifOpenService: IfOpenService,
+    private ifOpenService: ClrPopoverToggleService,
```
Wallabeng: lets also log a console.warn as deprecation message

Wallabeng:
Some more findings, after testing it locally:

- Cards are losing their shadow when hovering
- Breadcrumb should bei 13px in font-size again, I think it's too small otherwise
- Notifications: Font-size of text should also be 14px again
- Numericfield: Increase font-size of unit to 13px again and add simple dev-demos for form support (template & reactive)
- SearchField: Magnifying glass/Delete icon are not in line with text and too narrow
- Readonly radio/checkboxes: The selected-tick is missing
- Letter-Avatar: Change font-size of large-avatar to 1.2rem
- Main-Nav-Group: Change size of caret-down icon to 10px again

Wallabeng:
> I like the idea of moving to css custom variables. But keep in mind, that this breaks compatibility with IE 11. Applications need to use a polyfill to reenalbe IE 11 support.

Applications which need to support IE11 need to call the ponyfill provided by clarity.
This is explained at the bottom of this page https://clarity.design/documentation/themes

PR #731, opened 2020-04-07
fix: color for PHS theme in different places

derkoe:
This fixes the color for the PHS theme for progress spinner, radio button group and probably other places.

### DIFF
--- a/src/clr-addons/themes/phs/variables.color.scss
+++ b/src/clr-addons/themes/phs/variables.color.scss
@@ -1,3 +1,5 @@
+$clr-color-hue: 183.2;
+
 :root {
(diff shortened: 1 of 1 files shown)


src/clr-addons/themes/phs/phs-theme.scss
```diff
@@ -1,3 +1,5 @@
+$clr-color-hue: 183.2;
```
cat-mandres, line 1: actually, shouldnt we move this into the variables.color file of the phs folder - this is just the file importing all other files

derkoe: moved that to variables.color

PR #750, opened 2020-04-16
NumericField: Add validator directives

Wallabeng:


src/clr-addons/numericfield/numeric-field-validators.ts
```diff
@@ -87,3 +88,31 @@ export class ClrNumericFieldValidators {
     return parseFloat(formattedNumber);
   }
 }
+
+@Directive({
+  selector: '[clrMaxNumericField]',
```
cat-mandres, line 93: wouldnt a shorter selector here make sense, atleast without "field" at the end? maybe just [clrMax] or [clrNumericMax]

PR #777, opened 2020-04-29
Fixed combobox optionbox bug + Improved signpost width on mobile

cat-mrialto:


src/clr-addons/combobox/combobox.ts
```diff
@@ -24,12 +25,16 @@ import {
...
   ClrLabel,
   ClrPopoverToggleService,
+  ɵb as RootDropdownService,
   ɵba as LayoutService,
   ɵbb as NgControlService,
   ɵbd as IfErrorService,
   ɵbc as ControlClassService,
   ɵe as POPOVER_HOST_ANCHOR,
   ɵz as ControlIdService,
+  ɵj as DROPDOWN_FOCUS_HANDLER_PROVIDER,
```
Wallabeng, line 35: pls add all of those imports to the tests, so we can be sure that we import the correct stuff (also the ones from options.ts

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clr-addons",
-  "version": "7.4.0",
+  "version": "7.4.1",
```
Wallabeng, line 3: pls raise version to 7.5.0, as I updated clarity dependency to 3.1.1
to avoid this in future, just make a separate version commit

PR #801, opened 2020-05-15
Add custom icons to clr addons and provide them via ClarityIcons API

cat-mandres:


cat-mandres:
This would be causing a breaking change - each client must remove the clr-icons.js from their angular.json file.

also not sure about the naming we should go for, open to suggestions

jerone:
The preview link on the [releases page](https://github.com/porscheinformatik/clarity-addons/releases/tag/v8.0.0) seems to link to the news page: https://porscheinformatik.github.io/clarity-addons/news

cat-mandres:
thanks for pointing that out, fixed it. However the link will only work after this PR is merged aswell to get the new website version online: https://github.com/porscheinformatik/clarity-addons/pull/805

PR #816, opened 2020-06-18
New component: history

tschambalamba:


src/clr-addons/history/history-pinned.ts
```diff
@@ -0,0 +1,52 @@
...
+    this.historyService.cookieSettings$.subscribe(settings => {
+      const setting = settings.find(setting => setting.username === this.username);
+      if (setting) {
+        this.active = setting.historyPinned;
+      }
+    });
+  }
+
+  ngOnDestroy(): void {
+    this.settingsSubscription.unsubscribe();
```
Wallabeng, line 46: subscription not assigned to this variable

src/clr-addons/history/history.scss
```diff
@@ -0,0 +1,21 @@
+@include exports('history.clr-addons') {
```
Wallabeng, line 1: this whole file should be merged with history-pinned, as it defines rules for the pinned component

src/clr-addons/history/history.service.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  addHistoryEntry(historyEntry: ClrHistoryModel): void {
+    this.removeFromHistory(historyEntry);
+    let history = this.getHistory(historyEntry.username, historyEntry.applicationName, historyEntry.tenantid);
+    /* add it as last element */
+    if (historyEntry.tenantid) {
+      historyEntry.tenantid = this.tenantPrefix + historyEntry.tenantid;
+    }
+    history.push(historyEntry);
+    /* only consider the last 4 bread crumbs per tenantid */
```
Wallabeng, line 33: should read as "... per tenant and application"

src/clr-addons/history/history.service.ts
```diff
@@ -0,0 +1,138 @@
...
+    );
+  }
+
+  private setCookie(name: string, content: string): void {
+    document.cookie =
+      name +
+      '=' +
+      content +
+      ';domain=' +
+      window.location.hostname.split('-').slice(-1) +
```
Wallabeng, line 133: won't work in production, as links look like that over there:
https://outputmanager.cross3.at/crossng-outputmanager/outputmanager

src/clr-addons/history/history.service.ts
```diff
@@ -0,0 +1,138 @@
...
+        .filter(cookie => {
+          return cookie.substring(0, name.length + 1) === `${name}=`;
+        })
+        .map(cookie => {
+          return JSON.parse(cookie.substring(name.length + 1));
+        })[0] || []
+    );
+  }
+
+  private setCookie(name: string, content: string): void {
```
Wallabeng, line 127: this will not only set the cookies of the current application, but will overwrite all

src/clr-addons/history/history.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Component, Input, OnInit } from '@angular/core';
+import { ClrHistoryModel } from './history-model.interface';
+import { ClrHistoryService } from './history.service';
+
+@Component({
+  selector: 'clr-history',
+  templateUrl: './history.html',
+})
+export class ClrHistory implements OnInit {
+  @Input('username') username: string;
```
Wallabeng, line 16: we follow clarity styleguides, so input variables should prefix 'clr' (clrUsername in this case)
https://github.com/vmware/clarity/blob/master/docs/CODING_GUIDELINES_ANGULAR.md

src/clr-addons/history/history.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(ClrHistory);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
```
Wallabeng, line 33: some more tests pls

src/clr-addons/history/history.service.ts
```diff
@@ -0,0 +1,138 @@
...
+          element.pageName === entry.pageName &&
+          element.context.indexOf(entry.context) === 0 &&
+          element.tenantid === this.tenantPrefix + entry.tenantid &&
+          element.username === entry.username
+        )
+    );
+    this.setHistory(history);
+  }
+
+  getHistoryPinned(username: string): boolean {
```
Wallabeng, line 86: i would remove this method and "initialize" the cookieSettings observable inside the constructor, so you just need to subscribe to that.

src/clr-addons/history/history-pinned.ts
```diff
@@ -0,0 +1,52 @@
...
+import { ClrHistoryModel } from './history-model.interface';
+import { ClrHistoryService } from './history.service';
+import { Subscription } from 'rxjs';
+
+@Component({
+  selector: 'clr-history-pinned',
+  templateUrl: './history-pinned.html',
+})
+export class ClrHistoryPinned implements OnInit, OnDestroy {
+  @Input('username') username: string;
```
Wallabeng, line 17: we follow clarity styleguides, so input variables should prefix 'clr' (clrUsername in this case)
https://github.com/vmware/clarity/blob/master/docs/CODING_GUIDELINES_ANGULAR.md

src/clr-addons/history/history.service.ts
```diff
@@ -0,0 +1,143 @@
...
+   */
+  addHistoryEntry(historyEntry: ClrHistoryModel): void {
+    this.removeFromHistory(historyEntry);
+    let history = this.getHistory(historyEntry.username, historyEntry.applicationName, historyEntry.tenantid);
+    /* add it as last element */
+    if (historyEntry.tenantid) {
+      historyEntry.tenantid = this.tenantPrefix + historyEntry.tenantid;
+    }
+    history.push(historyEntry);
+    /* only consider the last 4 bread crumbs per tenantid */
```
Wallabeng, line 37: comment still not quite correct :)

src/clr-addons/history/history-model.interface.ts
```diff
@@ -0,0 +1,54 @@
...
+export interface ClrHistoryModel {
+  /**
+   * The username
+   */
+  username: string;
+
+  /**
+   * The tenant id
+   */
+  tenantid?: string;
```
Wallabeng, line 16: hmm just thought that "tenantId" and "applicationName" are very opinionated for a general component.
do you think we can refactor that to a context array?
and also include the existing context variable into it, as it seems it's just there to be able to add the same page multiple times with different path variables

PR #846, opened 2020-07-27
Fix layout for multilingual input/textarea

derkoe:
* use clr-textarea-wrapper for textares
* add margin to selector to get rid of it when hidden


derkoe:
This fixes:
1. the layout of the multilingual textarea (red mark)
2. the margin of the multilingual component when the selector is hidden (yellow mark)

image

PR #859, opened 2020-08-05
fix(history): use links in pinned history, fix dropdown position

derkoe:
- Pinned history should use default links (using href).
- History dropdown menu should have a configurable position with 'bottom-right' as the default
- Only display 3 entries of history
- Fix hideLast was of type "true"

### DIFF
--- a/src/clr-addons/history/history-pinned.html
+++ b/src/clr-addons/history/history-pinned.html
@@ -2,7 +2,7 @@
   <ol class="history">
     <ng-container *ngFor="let historyItem of historyElements">
       <li *ngIf="historyItem.url" class="history-item">
-        <a (click)="select(historyItem)">{{historyItem.title}}</a>
+        <a [href]="historyItem.url">{{historyItem.title}}</a>
       </li>
     </ng-container>
   </ol>
--- a/src/dev/src/app/app.component.ts
+++ b/src/dev/src/app/app.component.ts
@@ -3,13 +3,14 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { Component, PLATFORM_ID, Inject } from '@angular/core';
+import { Component, PLATFORM_ID, Inject, OnInit } from '@angular/core';
 import { Route } from '@angular/router';
 import { APP_ROUTES } from './app.routing';
 import { isPlatformBrowser, DOCUMENT } from '@angular/common';
+import { RouteHistoryService } from './route-history.service';
 
 @Component({ selector: 'app-root', templateUrl: './app.component.html' })
-export class AppComponent {
+export class AppComponent implements OnInit {
   public routes: Route[] = APP_ROUTES;
   linkRef: HTMLLinkElement;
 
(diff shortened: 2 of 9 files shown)


derkoe:
@d-m-s this is how it looks now.

image

src/dev/src/app/app.component.ts
```diff
@@ -21,13 +22,15 @@ export class AppComponent {
...
-    @Inject(PLATFORM_ID) private platformId: Record<string, any>
+    @Inject(PLATFORM_ID) private platformId: Record<string, any>,
+    private routeHistoryService: RouteHistoryService
   ) {
     if (isPlatformBrowser(this.platformId)) {
       this.linkRef = this.document.createElement('link');
       this.linkRef.rel = 'stylesheet';
       this.linkRef.href = this.themes[0].href;
       this.document.querySelector('head').appendChild(this.linkRef);
+      this.routeHistoryService.toString(); // needed for angular to create service instance
```
Wallabeng, line 33: this looks very weird... pls either copy the code from the service to app.component.ts or don't do it in constructor and call an "init" method or anything, as constructors shouldn't have any logic inside (according to best practices)

derkoe: what about AppComponent's constructor - seems to have some logic inside ;-)

derkoe: Changed in latest commit

Wallabeng: one doesn't know best practices when he starts something new :-)

src/clr-addons/history/history-pinned.html
```diff
@@ -2,7 +2,7 @@
   <ol class="history">
     <ng-container *ngFor="let historyItem of historyElements">
       <li *ngIf="historyItem.url" class="history-item">
-        <a (click)="select(historyItem)">{{historyItem.title}}</a>
+        <a [href]="historyItem.url">{{historyItem.title}}</a>
```
Wallabeng, line 5: sorry forgot in last review: initially (in menu bar also) we did redirects with "location.href = ..." because IOS would lose "app mode" (when switching applications), when "installing" our apps on home screen to access it as "PWA".
so when changing this to href, we need to check this first (@d-m-s also knows this topic)

derkoe: @d-m-s requested this change - what should we do?

d-m-s: 1. **we need to ensure proper link styling!** 
valid "a href" seems to be required for clarity to apply styling.

1. **from ux-point of view we should enable the user to open a history-item in a new tab quickly** 
("CRTL+Click", "Left-Click > open in new Tab", ...)

1. **App-Mode must be supported and UX must not be brocken on mobile devices.**

d-m-s: discussed with @derkoe, we'll go with the current commit and need to add some fixing for the app-mode in #860. 
Probably there is some possibility to prevent the app-mode from quitting when referring to another tld-url.

PR #897, opened 2020-09-11
chore(deps): switch to Clarity 4 and Angular 10, remove combobox

### LINKED ISSUE #823: Angular 10 / Clarity 4 Update
Update to Clarity 4 and Angular 10 and possibly provide `.next` builds like Clarity does before the release.

derkoe:
- Update relase to 9.0.0
- Update to Clarity 4.0.1
- Update to Angular 10.0.14 (cannot update to 10.1.x because of https://github.com/vmware/clarity/issues/5015)
- Remove deprecated combobox (because of name clash with new https://clarity.design/documentation/combobox)

Closes #823


package.json
```diff
@@ -33,31 +33,31 @@
...
+    "@angular/animations": "10.0.14",
+    "@angular/common": "10.0.14",
+    "@angular/compiler": "10.0.14",
+    "@angular/core": "10.0.14",
+    "@angular/forms": "10.0.14",
+    "@angular/platform-browser": "10.0.14",
+    "@angular/platform-browser-dynamic": "10.0.14",
+    "@angular/router": "10.0.14",
+    "@clr/angular": "4.0.1",
     "@clr/core": "3.1.5",
```
Wallabeng, line 45: this also needs an update... did you update by hand or with "ng update..."? later one is recommended, so the migration scripts of angular and clarity get executed. so if you did by hand, pls run those scripts to be sure.

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compileOnSave": false,
   "compilerOptions": {
     "importHelpers": true,
-    "module": "esnext",
+    "module": "es2020",
```
Wallabeng, line 5: is there something wrong with esnext? as clarity also compiles that way, i would recommend keeping it

Wallabeng (changes requested):

package.json
derkoe: Thx - forgot this again. 

I did use `ng update`

tsconfig.json
derkoe: `ng update` did this

derkoe: https://angular.io/guide/migration-update-module-and-target-compiler-options

package.json
derkoe: also forgot to adapt the package.json in npm/clr-addons - will fix that

derkoe: Fixed in current commit

Wallabeng (approved):

PR #918, opened 2020-09-25
Add size constraints to clr.history cookie handling

steell0815:


src/clr-addons/history/history.service.ts
```diff
@@ -28,12 +33,15 @@ export class ClrHistoryService {
...
     this.removeFromHistory(historyEntry);
     let history = this.getHistory(historyEntry.username, historyEntry.context);
-    /* add it as last element */
-    history.push(historyEntry);
 
-    /* support a maximum of 4 pages in history */
-    history = history.slice(-4);
-    this.setHistory(history, historyEntry.username, domain);
+    /* only add to history in case url does not exceed maxUrlSize characters */
+    if (historyEntry.url && historyEntry.url.length < this.maxUrlSize) {
```
Wallabeng, line 38: Can we return false, if the url is not added to the history? Otherwise I bet there will be confusion

Wallabeng (changes requested):

Wallabeng (approved):

PR #954, opened 2020-11-03
Fix content panel wrapper z-index

magistvan:


src/clr-addons/content-panel/content-panel.scss
```diff
@@ -33,7 +33,7 @@
 
   .content-panel-content-wrapper {
     background-color: var(--clr-global-app-background);
-    z-index: map-get($clr-layers, dropdown-menu) - 1;
+    z-index: map-get($clr-layers, sidepanel-bg) - 1;
```
cat-mandres, line 36: also adapt the z-index of the content-pnale-backdrop, line 82

cat-mandres (approved):

PR #1071, opened 2021-05-04
fix: breadcrumbs/history encode/decode fixing

tsschnabl:
breadcrumbs (history) has a problem on stringify non-ascii UTF-8 characters in the url.
uriEncode the value first solves the issue (and uriDecode on loading)

### DIFF
--- a/src/clr-addons/history/history.service.ts
+++ b/src/clr-addons/history/history.service.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2020 Porsche Informatik. All Rights Reserved.
+ * Copyright (c) 2018-2021 Porsche Informatik. All Rights Reserved.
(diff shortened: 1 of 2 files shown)


src/clr-addons/history/history.service.ts
```diff
@@ -188,14 +188,14 @@ export class ClrHistoryService {
   }
 
   private encode(content: ClrHistoryModel[]): string {
-    const jsonString = btoa(JSON.stringify(content));
+    const jsonString = btoa(encodeURIComponent(JSON.stringify(content)));
```
Wallabeng, line 191: pls add a test to check this behaviour, so we don't run into it again in the future

Wallabeng (changes requested):

tsschnabl:
I've added the test

PR #1131, opened 2021-05-26
update clarity-addons version for new release

ExtMuthM:


tsschnabl (commented):
Thx to create a version, so we can use the history breadcrumb fix

Wallabeng (approved):

PR #1183, opened 2021-06-24
fix ObjectUnsubscribedError if the component is destroyed quickly

daWieser:


src/clr-addons/generic-quick-list/generic-quick-list.ts
```diff
@@ -67,6 +70,9 @@ export class ClrGenericQuickList<T extends ClrGenericQuickListItem> implements O
   }
 
   setFocusOnAdd(): void {
+    if (this.destroyed)
```
Wallabeng, line 73: Could you instead please remember the return of setTimeout and use clearTimeout inside ngOnDestroy?
Would be more future proof if any method gets added inside setTimeout

Wallabeng (changes requested):

Wallabeng (approved):

PR #1211, opened 2021-07-15
Upgrade Clarity to 5.2.0

SzVeress:
+ upgrade to clarity 5.2.0
+ load newly added (CDS) Timeline icon
+ fix use of "whitelistedNonPeerDependencies" being deprecated

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "cpy-cli": "3.1.1",
     "cross-var": "1.1.0",
     "csso-cli": "3.0.0",
-    "del-cli": "4.0.0",
+    "del-cli": "4.0.1",
     "eslint": "7.30.0",
     "eslint-config-prettier": "8.3.0",
     "eslint-plugin-jasmine": "4.1.2",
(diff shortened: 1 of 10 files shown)


src/clr-addons/icons/cds-icons.ts
```diff
@@ -0,0 +1,6 @@
+import { ClarityIcons } from '@cds/core/icon';
+import { timelineIcon } from '@cds/core/icon/shapes/timeline';
+
+export function initCdsIcons(): void {
+  ClarityIcons.addIcons(timelineIcon);
+}
```
Wallabeng, line 6: We don't want to add icons into every bundle which uses clarity-addons. So if you need cds icons pls import them in your specific project (or better in the specific module inside the project, to keep bundle sizes small).

Wallabeng (changes requested):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clr-addons",
-  "version": "10.1.9",
+  "version": "10.1.10",
```
Wallabeng, line 3: Raising clarity versions is at least a minor release, because that version also needs to be updated in client code.
Please raise to 10.2.0

Wallabeng (changes requested):

Wallabeng (approved):

PR #1412, opened 2021-12-07
Fix clr-combobox-wrapper and clr-combobox-input width

zsoltx:


martinbrom:
Increase the versions too

PR #1684, opened 2022-05-29
Add clarity-icons to 10.x.x

steell0815:


steell0815:
Hi @Wallabeng, I would need the new icons in a new release 10.5.2, being compatible to angular 11. However, I am still working on angular 13 upgrade for service-base (see https://gerrit-review.porscheinformatik.com/c/products/crossng/service-base/+/46896, but reverted after unsuccessful tests on CD) is already in progress. Maybe I can fix it by 30th May 2022.

In case not, I would perform the upgrade in R.1.67 but update the icons in R.1.66.

Many thanks, Stefan

PR #1719, opened 2022-06-28
added rows directive to textarea

danieldeusing:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clr-addons",
-  "version": "12.5.2",
+  "version": "12.6.2",
```
Wallabeng, line 3: pls use 12.6.0

src/clr-addons/multilingual/abstract-multilingual.ts
```diff
@@ -14,6 +14,7 @@ export abstract class ClrMultilingualAbstract extends ClrAbstractFormComponent {
   @Input('clrSelectedLang') selectedLang: string;
   @Input() readonly: string;
   @Input() maxlength: number;
+  @Input() rows: number;
```
Wallabeng, line 17: pls move this to multilingual-textarea component, as it's not used in single line input

Wallabeng (changes requested):

package.json
danieldeusing: Bad mistake, done

src/clr-addons/multilingual/abstract-multilingual.ts
danieldeusing: done

Wallabeng (approved):

PR #1749, opened 2022-08-23
Rename icon due to name collision with Clarity

ueberp:
An icon with the name dollar-bill already exists in Clarity. This leads
to a name collision, possibly resulting in the wrong icon to be
displayed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clr-addons",
-  "version": "12.8.1",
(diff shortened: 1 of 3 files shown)


ueberp:
An icon named 'dollar-bill' is present in both icon sets.
https://porscheinformatik.github.io/clarity-addons/icons/icon-sets
https://clarity.design/icons
This change renames the icon in clarity-addons.

PR #1753, opened 2022-09-13
Added new clarity icon business partner with car needed for C3DEV-37271

Pamando:


martinbrom (changes requested):
Could you please increase the patch version also in the root package.json? Also, the package-lock says .2 and the package.json .3 😉

martinbrom (approved):

PR #1766, opened 2022-09-28
chore: integrate the website into the master branch

derkoe:


.github/workflows/publish.yml
```diff
@@ -6,14 +6,28 @@ jobs:
...
+      - name: Use Node.js 16.x
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+      - name: Build
+        run: |
+          npm ci
+          npm run build
+      - name: Publish
+        run: npm run publish
```
Wallabeng, line 19: could you also create a workflow to publish the website on its own?

Wallabeng: which should only be triggered manually I guess, because now we can block the main commit without demo code. so website-only-deploys shouldn't happen that often.

Wallabeng (approved):

PR #1826, opened 2022-11-02
Introduce semver for clarity dependencies

### LINKED ISSUE #1821: Loosen versions of Clarity
Would it be possible to loosen the supported versions of the Clarity dependencies?

Right now, those dependencies are [hard-coded](https://github.com/porscheinformatik/clarity-addons/blob/0136ba6f9061d135c3c6a4f9869e9d121fe8bac5/src/clr-addons/package.json#L23-L26), even on semver patch level.

I understand that `@clr/angular` & `@clr/icons` & `@clr/ui` are strong dependencies within this project. And that there is code depending on features of those dependencies.
Can I suggest adding tilde symbol `~` to get at least every update within the same minor range, because semver patch level should not include any breaking changes. 

For `@cds/core` I want to suggest adding an range of supported versions: `^5.6.0 || ^6.0.0`.
To keep in sync with what [Clarity suggests themselves](https://github.com/vmware-clarity/ng-clarity/blob/89858ae183837bbbf8a8241588db509bd09ee8fa/projects/angular/package.json#L23).

The reason for this is to get more control over what versions of Clarity we can use. And that we don't have to wait for Clarity Addons to support those versions.

image

Thank you, keep up the work on adding Addons!

Wallabeng:
closes #1821


jerone (approved):
Yes, this allows us to update patch version of `@clr/*` packages and minor versions of `@cds/core`. 👍

PR #1851, opened 2022-11-28
Datagrid state persistence directive

Wallabeng:


src/clr-addons/datagrid/datagrid-state-persistence/state-persistence-key.directive.ts
```diff
@@ -1,35 +1,156 @@
...
+const DATE_TYPE = 'date';
 
 @Directive({
   selector: '[clrStatePersistenceKey]',
 })
-export class StatePersistenceKeyDirective implements AfterContentInit {
-  @Input()
-  clrStatePersistenceKey: string;
+export class StatePersistenceKeyDirective implements AfterContentInit, OnDestroy {
+  @Input('clrStatePersistenceKey')
```
martinbrom, line 13: could we skip the breaking change here? by using another input and hardcoding a default value for the `serverDriven` ?

src/clr-addons/datagrid/datagrid-state-persistence/state-persistence-key.directive.ts
```diff
@@ -1,35 +1,156 @@
...
+      Object.keys(state.columns).forEach(prop => (state.columns[prop].filterValue = undefined));
+      dgState.filters?.forEach(filter => {
+        const property = this.getFilterPropertyName(filter);
+        if (property) {
+          state.columns[property] = state.columns[property] || {};
+          state.columns[property].filterValue = this.enrichFilterValue(filter);
+        }
+      });
+
+      sessionStorage.setItem(this.options.key, JSON.stringify(state));
```
martinbrom, line 99: see Teams message about `LocalStorage` vs `SessionStorage`

src/dev/src/app/datagrid-state-persistence/datagrid-state-persistence.demo.ts
```diff
@@ -1,7 +1,25 @@
...
+    [...Array(15).keys()].map(i => ({
+      hideableCol: 'item' + i,
+      numericCol: i,
+      dateCol: new Date(),
+      enumCol: 'Enum ' + i,
+    }))
+  ).pipe(delay(0));
+
+  refresh(state: ClrDatagridStateInterface) {
+    console.log('data refresh needed', JSON.stringify(state));
```
martinbrom, line 23: intentional or forgotten here?

src/clr-addons/datagrid/datagrid-state-persistence/state-persistence-key.directive.ts
```diff
@@ -1,35 +1,156 @@
...
+  getLocalStorageState(): ClrDatagridStatePersistenceModel {
+    return JSON.parse(localStorage.getItem(this.options.key)) || {};
+  }
+
+  /**
+   * As a date is serialized as string, but not deserialized as date
+   * we need to add some meta information to do that manually later
+   */
+  private enrichFilterValue(filter: any) {
+    const result = {} as any;
```
martinbrom, line 116: we're gonna be returning any here, which is not ideal. Could we use a `Record` type here as the return value? like `Record<keyof filter, unknown>`

martinbrom (changes requested):
I also left a few other general questions in Teams, would be nice if we can also discuss those

src/clr-addons/datagrid/datagrid-state-persistence/state-persistence-key.directive.ts
Wallabeng: I also thought of that, but because of the drawbacks stated in the documentation I want the users to actively decide what they are doing

src/clr-addons/datagrid/datagrid-state-persistence/state-persistence-key.directive.ts
Wallabeng: volatile values (current page and filters) are stored in SessionStorage to not have the need of a reset button.
more static values (hidden state and page size) are stored in LocalStorage as they should almost always stay the same.

src/dev/src/app/datagrid-state-persistence/datagrid-state-persistence.demo.ts
Wallabeng: intentional, because as stated in the documentation there are drawbacks and especially when you use a server filtered datagrid you need special attention on how often this method is called.

src/clr-addons/datagrid/datagrid-state-persistence/state-persistence-key.directive.ts
Wallabeng: yes oversaw that, will remove the any

Wallabeng: done

src/clr-addons/datagrid/datagrid-state-persistence/state-persistence-key.directive.ts
martinbrom: option added, resolved

martinbrom (approved):
lgtm, great addition :)

PR #1861, opened 2022-12-19
Fix history service

fafeichter:
This error was the reason why pages like user settings under certain circumstances are not loaded on 1. click.

https://jira.porscheinformatik.com/jira/browse/C3DEV-48733


fafeichter:
@Wallabeng

Wallabeng (approved):

PR #1871, opened 2023-01-27
Forward keyup event in components for multilingual inputs

fafeichter:
Parents should be able to listen for the keyup events from the `<input>` element.

This is required for [C3DEV-45925](https://gerrit-review.porscheinformatik.com/c/products/crossng/service-base/+/73327).


Wallabeng (changes requested):
Hmm currently I don't see the need for this event to be forwarded, as you get the same results when using [ngModelOptions]="{ updateOn: 'change' }" or just remove the existing ngModelOptions as 'change' is the default

fafeichter:
I replaced `[ngModelOptions]` and `(keyup)` with `(ngModelChange)="updateModel($event)"` which makes this change unnecessary.

PR #1885, opened 2023-02-14
update campaign and campaign-outdated icons

igorradovanovic:


steell0815:
pls fix the verifier failure

PR #1930, opened 2023-08-01
add state colors in datagrid and notification, update documentation

tschambalamba:


src/clr-addons/themes/phs/css.overrides.scss
```diff
@@ -63,6 +63,84 @@
...
+    ):hover {
+    background-color: $clr-highlight-neutral-200;
+
+    .datagrid-select {
+      background-color: $clr-highlight-neutral-200;
+    }
+  }
+
+  .btn-success {
+    background-color: $clr-highlight-success-700;
```
Wallabeng, line 95: instead of overriding css for that, we could just rewrite the css3 variables already in use by the buttons.
e.g. btn-success uses the variable --clr-btn-success-bg-color, which should be overwritten by the PHS theme instead

src/clr-addons/treetable/treetable.scss
```diff
@@ -241,7 +242,7 @@
         width: auto;
 
         .treetable-row:hover {
-          background-color: var(--clr-datagrid-row-hover);
+          background-color: $clr-highlight-neutral-200;
```
Wallabeng, line 245: same here just use the variable --clr-datagrid-row-hover-color and overwrite it's value to this

Wallabeng (changes requested):

Wallabeng (approved):

PR #1932, opened 2023-08-02
add crlSize to brand avatar, change to cds-icon

tschambalamba:


src/clr-addons/brand-avatar/brand-avatar.ts
```diff
@@ -25,16 +48,31 @@ export class ClrBrandAvatar {
...
+  defaultIcon = false;
+
+  constructor(private renderer: Renderer2) {}
+
+  ngAfterViewInit(): void {
+    this.renderer.setStyle(this.avatarElement.nativeElement, 'width', this.size + 'px');
+    this.renderer.setStyle(this.avatarElement.nativeElement, 'height', this.size + 'px');
+  }
+
+  ngAfterContentInit(): void {
```
Wallabeng, line 66: i guess this should be done more frequent, either in a getDefaultIcon method or when "clrBrand" has been set. I refer to async loading of data when the brand comes in at any time

Wallabeng (changes requested):

PR #1962, opened 2023-09-25
Datagrid enum filter: Add custom display name

lefkas95:
We would like to define a "display name" for enums in the `enum-filter` component. The reason is that the name of an enum in the code is often not user friendly. Also, for i18n purposes such a functionality is very useful for us.

Therefore, we extended the `clrPossibleValues` to either accept a list of strings (current master behaviour) or a list of `{ value: string, displayValue: string }` objects.

The demo project was extended to showcase our change:

**Current behaviour**
Enum values are shown as they are defined in code - not very user friendly
image

**Extension to specify a display name**
image

### DIFF
--- a/src/clr-addons/datagrid/enum-filter/enum-filter.component.ts
+++ b/src/clr-addons/datagrid/enum-filter/enum-filter.component.ts
@@ -5,6 +5,11 @@ import { ClarityIcons, trashIcon } from '@cds/core/icon';
 
 ClarityIcons.addIcons(trashIcon);
 
+interface FilterValue {
+  value: string;
+  displayValue: string;
+}
+
 @Component({
   selector: 'clr-enum-filter',
   templateUrl: './enum-filter.component.html',
@@ -17,25 +22,26 @@ export class ClrEnumFilterComponent<T extends { [key: string]: any }>
 
   @Input('clrFilterValues')
   public set value(values: string[]) {
+    const converted = values.map(filtered => ({ value: filtered, displayValue: filtered }));
     if (this.possibleValues?.length) {
-      this.filteredValues = values.filter(filtered => this.possibleValues.includes(filtered));
-      this.clrFilterValuesChange.emit(this.filteredValues);
+      this.filteredValues = converted.filter(filtered => this.possibleValues.includes(filtered));
+      this.clrFilterValuesChange.emit(this.getDisplayValues(this.filteredValues));
     } else {
-      this.filteredValues = values;
+      this.filteredValues = converted;
     }
     this.changes.emit(true);
   }
 
   @Output() clrFilterValuesChange = new EventEmitter<string[]>();
 
-  possibleValues: string[] = [];
+  possibleValues: FilterValue[] = [];
   customPossibleValues = false;
-  filteredValues: string[] = [];
+  filteredValues: FilterValue[] = [];
   changes = new EventEmitter<boolean>(false);
 
   destroyed$ = new Subject<void>();
 
-  @Input() set clrPossibleValues(values: string[]) {
+  @Input() set clrPossibleValues(values: (string | FilterValue)[]) {
     this.setPossibleValues(values);
     this.customPossibleValues = true;
   }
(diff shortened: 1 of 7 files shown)


Wallabeng:
Hi @lefkas95,

first of all thank you for the contribution.
From UX perspective I'm not quite sure if it makes sense to show a different value inside the filter than in the datagrid itself.
As new you would filter for "Readable Value 1", but in datagrid this means "TestValue1" which seems quite confusing.

If we need translated values, we translate before passing the values to the datagrid.
So the translated value will be shown in filter and datagrid cell and the enum filter works out of the box.

Do you have a different usecase for that?

d-m-s:
I do agree, from ux point of view the values in the filter-overlay need to be the same as the data-values in the column.
The mentioned approach from @Wallabeng should be a good solution for your problem.

jerone:
I second this feature request. For the same reason I actually build my own filter.

We do not always control the data from the server. 
Or it's to complex to alter the data.
Or we cannot control/alter the enum.
It's very often we show different text/icons then the enum values.

lefkas95:
Thank you for your fast feedback. In our case, the data source is our REST backend and we use the output `(clrDgRefresh)` of `clr-datagrid` to get and append the filters to our REST request. As we i18n the enums (so the users can actually read them) that would mean, the enum filter in `ClrDatagridStateInterface` would contain the translated enum and we would, at some point, need to map the translated enum back to its actual value.

This is not very optimal - what do you think?

Wallabeng:
Alright I see your point.
Could you then please also update the documentation (folder "website"), with following information:
- Usage of your implementation
- When to use (basically only for server side filtering I guess)
- Generally the datagrid cell "display value" should be the same as the filter "display value", as a UX guideline (if you want to use it differently in your projects, feel free 😄)

lefkas95:
I updated my solution and cleaned it up a bit. The fields `possibleValues` and `filteredValues` are now always of type `FilterValue` (`{ value: string; displayValue: string; }`) to get rid of those "is it a string or object" checks. If the input `clrPossibleValues` is called with a string array, I simply convert them to `FilterValue`.

When is which property used?
- `displayValue`: For filtering the items and for showing the labels
- `value`: Passed to the `state` object to build the filter object

I updated the documentation and now also the same values are shown in the datagrid and the filter.
I tested this with various combinations and found no issues. Looking forward for feedback :)

d-m-s:
@lefkas95 @Wallabeng 
Please make sure that the following things are mentioned in the documentation:

1. UX Guideline "Recognition: Displayed values in the filter-overlay need to be the same as the data-values in the column."
2. Document a code pattern/best practice how to implement it normally
3. The workaround should only be used if 2.) is not possible and you have to take care yourself that 1.) is respected.

lefkas95:
Documentation extended and all 3 points are mentioned there

src/clr-addons/datagrid/enum-filter/enum-filter.component.ts
```diff
@@ -17,25 +22,26 @@ export class ClrEnumFilterComponent<T extends { [key: string]: any }>
 
   @Input('clrFilterValues')
   public set value(values: string[]) {
+    const converted = values.map(filtered => ({ value: filtered, displayValue: filtered }));
     if (this.possibleValues?.length) {
-      this.filteredValues = values.filter(filtered => this.possibleValues.includes(filtered));
-      this.clrFilterValuesChange.emit(this.filteredValues);
+      this.filteredValues = converted.filter(filtered => this.possibleValues.includes(filtered));
+      this.clrFilterValuesChange.emit(values);
```
Wallabeng, line 28: this should only contain values from "this.filteredValues", not all given as input parameters

src/clr-addons/datagrid/enum-filter/enum-filter.component.ts
```diff
@@ -51,19 +57,21 @@ export class ClrEnumFilterComponent<T extends { [key: string]: any }>
     });
   }
 
-  setPossibleValues(values: string[]) {
-    this.possibleValues = values;
+  setPossibleValues(values: (string | FilterValue)[]) {
+    this.possibleValues = values.map(v => (v instanceof Object ? v : { value: v, displayValue: v }));
     this.possibleValues.sort();
```
Wallabeng, line 62: this should sort the values by displayValue

src/clr-addons/datagrid/enum-filter/enum-filter.component.ts
```diff
@@ -90,12 +98,24 @@ export class ClrEnumFilterComponent<T extends { [key: string]: any }>
...
     this.changes.emit(true);
   }
 
   ngOnDestroy() {
     this.destroyed$.next();
     this.destroyed$.complete();
   }
+
+  protected containsFilterValue(entries: FilterValue[], filtered: FilterValue): boolean {
+    return !!entries.find(v => v.value === filtered.value);
```
Wallabeng, line 111: you could use "entries.some(..)" which directly returns a boolean

Wallabeng (changes requested):

src/clr-addons/datagrid/enum-filter/enum-filter.component.ts
lefkas95: done

src/clr-addons/datagrid/enum-filter/enum-filter.component.ts
lefkas95: adapted sorting

src/clr-addons/datagrid/enum-filter/enum-filter.component.ts
lefkas95: changed

Wallabeng (approved):

PR #1973, opened 2023-10-05
Fixes for asynchronously resolved nested dropdowns

martinbrom:
- fix styles not being applied after a nested dropdown is displayed and hidden again
- fix multiple subscriptions to ContentChildren changes


martinbrom:
Marcel FYI, opted to not increase a version as this currently doesn't affect any of our apps (and even if it would the impact is quite low unless there are tons of async emissions/changes). So I'd propose it's fine to be released with the next patch 👍

Wallabeng (approved):

PR #1984, opened 2023-10-13
Fix padding for .content-header on mobile devices

martinbrom:
just something small I noticed, no ticket so no version needed


martinbrom:
As the rule was completely overwritten by the base CSS because of higher specificity, the only difference is that on SM and smaller the padding-left is now `.5rem` instead of `1rem` (to be aligned with the padding-left on the `.content-area`

PR #2054, opened 2024-01-02
Persist hidden state independently

lefkas95:
Until now, `clrDgField` directive was needed to persist the hidden-state of a column. We would like to extend persisting the hidden-state by defining a persistance key in the directive. This way, one can persist the hidden-state of a column without the need to set `clrDgField`. So `clrDgField` still remains the main source for the persistance key and fallbacks to the implemented extension.

Background is, that we have a couple of columns with calculated values and a server-driven datasource, therefore filtering is not possible on those columns. Still, we want to persist the hidden-state here.

I tried out different approaches and prefered the presented one here as it is slim and non-inversive (as the existing `clrDgHideableColumn` input is simply extended by another property), but I'm open for remarks or alternative ideas. I wanted to wait with extending the documentation until first feedback.

### DIFF
--- a/src/clr-addons/datagrid/datagrid-state-persistence/column-hidden-state-persistence.directive.ts
+++ b/src/clr-addons/datagrid/datagrid-state-persistence/column-hidden-state-persistence.directive.ts
@@ -21,7 +21,7 @@ export class ColumnHiddenStatePersistenceDirective implements OnInit, OnDestroy 
 
   ngOnInit() {
     const persistenceEnabled = this.statePersistenceKey?.options?.persistHiddenColumns ?? true;
-    if (this.statePersistenceKey?.options.key && this.columnDirective?.clrDgField && persistenceEnabled) {
+    if (this.statePersistenceKey?.options.key && this.columnDirective?.persistenceKey && persistenceEnabled) {
       /* set hidden states from local storage (if existing) */
       this.initHiddenState();
 
@@ -37,9 +37,9 @@ export class ColumnHiddenStatePersistenceDirective implements OnInit, OnDestroy 
     const persistedGridState = this.readStoredState();
 
     /* read column state if existing */
-    if (persistedGridState?.columns?.[this.columnDirective.clrDgField]) {
+    if (persistedGridState?.columns?.[this.columnDirective.persistenceKey]) {
       /* read column hidden state if existing */
-      const persistedColumnHiddenState = persistedGridState.columns[this.columnDirective.clrDgField].hidden;
+      const persistedColumnHiddenState = persistedGridState.columns[this.columnDirective.persistenceKey].hidden;
       if (persistedColumnHiddenState !== undefined) {
         this.hideableColumnDirective.clrDgHidden = persistedColumnHiddenState === true;
       }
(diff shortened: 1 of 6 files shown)


martinbrom:
Hi @lefkas95, from my point of view this change would be okay :) Is it fine for you to wait until Monday when @Wallabeng is back in the office for another opinion?

Could you also please adapt the public docs in the `website` directory? A small mention of this new feature would surely be appreciated by others. Thank you so far!

src/clr-addons/datagrid/datagrid-state-persistence/column-hidden-state-persistence.directive.ts
```diff
@@ -10,6 +10,8 @@ import { takeUntil } from 'rxjs/operators';
   selector: '[clrDgHideableColumn]',
 })
 export class ColumnHiddenStatePersistenceDirective implements OnInit, OnDestroy {
+  @Input()
+  clrDgHideableColumn?: { hidden?: boolean; key?: string };
```
Wallabeng, line 14: I'm not that happy to extend clrDgHideableColumn as this is a clarity directive, not clarity-addons.
I would suggest to set the fieldKey with a new directive, which should be placed to clr-dg-column, so we may use it for other purposes too.
So basically you can extend "DatagridFieldDirective" with a different selector to not match the clarity one and provide a getter which returns either the one or the other and use this getter here instead of "this.columnDirective.clrDgField"

Wallabeng (changes requested):

src/clr-addons/datagrid/datagrid-state-persistence/column-hidden-state-persistence.directive.ts
lefkas95: Thank you for your feedback. I also had this concern a bit and changed it now according to your suggestion. `DatagridFieldDirective` has now a second selector and provides a getter for the "persistence" key.

Also updated the documentation a bit.

src/clr-addons/datagrid/datagrid-state-persistence/datagrid-field.directive.ts
```diff
@@ -1,9 +1,15 @@
...
-  selector: '[clrDgField]',
+  selector: '[clrDgField],[clrDgFieldKey]',
 })
 export class DatagridFieldDirective {
   @Input()
   clrDgField: string;
+  @Input()
+  clrDgFieldKey: string;
+
+  get persistenceKey(): string {
```
Wallabeng, line 12: could you refactor this getter to a method (getFieldName) pls?
personally I don't like getter with logic inside as this is so magic when you use it (you think it is a property but actually calculates something).

Wallabeng (changes requested):

src/clr-addons/datagrid/datagrid-state-persistence/column-hidden-state-persistence.directive.ts
Wallabeng: this looks good now thanks!
maybe just a small hint for future development in github (and other git repositories with merge/pull requests): just add new commits to the development branch.
because if you use amend and force push, we as reviewers can't see what the latest changes since the last review were and the discussions don't have reference points anymore

lefkas95: Yes sure, I will avoid ammending commits :)

src/clr-addons/datagrid/datagrid-state-persistence/datagrid-field.directive.ts
lefkas95: Changed it

Wallabeng (approved):

PR #2063, opened 2024-01-11
Feature/add page size options in result list

zsoltx:


martinbrom:
image

(as seen in Magellan already)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clr-addons",
-  "version": "15.7.0",
+  "version": "15.7.2",
```
martinbrom, line 3: I'd probably go for 15.8.0 here, but you don't have to increase the versions, we can release it in a separate step

src/clr-addons/paged-search-result-list/paged-search-result-list.html
```diff
@@ -1,23 +1,44 @@
...
-    (clrPageChange)="pageChange.emit($event)"
-  ></clr-pager>
+<div *ngIf="clrPagerPosition === 'bottom' || clrPagerPosition === 'both'">
+  <ng-container *ngTemplateOutlet="pager"></ng-container>
 </div>
+<ng-template #pager>
+  <div class="clr-row">
+    <div
+      [class.pager-top]="clrPagerPosition === 'top' || clrPagerPosition === 'both'"
+      [class.pager-bottom]="clrPagerPosition === 'bottom' || clrPagerPosition === 'both'"
```
Wallabeng, line 14: pager-top and pager-bottom should not depend on the provided options, but on the location where the pager is shown.
because now if you set 'both' the pager at the top also gets the class 'pager-bottom', right?

Wallabeng (changes requested):

src/clr-addons/paged-search-result-list/paged-search-result-list.html
```diff
@@ -1,23 +1,44 @@
...
-    [clrTotalItems]="totalItems"
-    (clrPageChange)="pageChange.emit($event)"
-  ></clr-pager>
+<div *ngIf="clrPagerPosition === 'bottom' || clrPagerPosition === 'both'">
+  <ng-container [ngTemplateOutlet]="pager" [ngTemplateOutletContext]="{ $implicit:'bottom' }"></ng-container>
 </div>
+<ng-template #pager let-position>
+  <div class="clr-row">
+    <div
+      [class.pager-top]="position === 'top' && (clrPagerPosition === 'top' || clrPagerPosition === 'both')"
```
Wallabeng, line 13: following checks are not needed right?
&& (clrPagerPosition === 'top' || clrPagerPosition === 'both')

src/clr-addons/paged-search-result-list/paged-search-result-list.html
```diff
@@ -1,23 +1,44 @@
...
+      [class.pager-bottom]="position === 'bottom' && (clrPagerPosition === 'bottom' || clrPagerPosition === 'both')"
```
Wallabeng, line 14: same here

Wallabeng (changes requested):

src/clr-addons/paged-search-result-list/paged-search-result-list.html
```diff
@@ -1,18 +1,23 @@
 <div *ngIf="clrPagerPosition === 'top' || clrPagerPosition === 'both'">
-  <ng-container [ngTemplateOutlet]="pager" [ngTemplateOutletContext]="{ $implicit:'top' }"></ng-container>
+  <ng-container
+    [ngTemplateOutlet]="pager"
+    [ngTemplateOutletContext]="{ $implicit: clrPagerPosition === 'top' || clrPagerPosition === 'both' ? 'top' : undefined }"
```
Wallabeng, line 4: why not just pass 'top'? as this instance here is always top

src/clr-addons/paged-search-result-list/paged-search-result-list.html
```diff
@@ -1,18 +1,23 @@
...
+  </ng-container>
 </div>
 <ng-container *ngFor="let item of items" [ngTemplateOutlet]="itemTemplate" [ngTemplateOutletContext]="{item: item}">
   <div *ngIf="!itemTemplate">Item template input is undefined!</div>
 </ng-container>
 <div *ngIf="clrPagerPosition === 'bottom' || clrPagerPosition === 'both'">
-  <ng-container [ngTemplateOutlet]="pager" [ngTemplateOutletContext]="{ $implicit:'bottom' }"></ng-container>
+  <ng-container
+    [ngTemplateOutlet]="pager"
+    [ngTemplateOutletContext]="{ $implicit: clrPagerPosition === 'bottom' || clrPagerPosition === 'both' ? 'bottom' : undefined }"
```
Wallabeng, line 14: why not just pass 'bottom'? as this instance here is always bottom

Wallabeng (changes requested):

src/clr-addons/paged-search-result-list/paged-search-result-list.html
zsoltx: true, as it already has an ngIf in the parent

Wallabeng (approved):

PR #2072, opened 2024-01-17
add css for custom alignment of the badge counter on filter

tschambalamba:


src/clr-addons/components.clr-addons.scss
```diff
@@ -522,5 +522,15 @@ clr-dropdown-menu .dropdown-header {
   min-height: 0.9rem;
 }
 
+//custom alignment of the badge counter on filter
+.clr-custom-filter-counter-badge {
+  position: absolute;
+  top: -7px;
+  right: -13px;
+  background-color: $turquoise;
```
Wallabeng, line 530: pls use --clr-badge-info-bg-color instead
or can't we just use an info badge here, so we don't need to specify the colors at all?

src/clr-addons/components.clr-addons.scss
```diff
@@ -522,5 +522,15 @@ clr-dropdown-menu .dropdown-header {
...
+  color: $white;
```
Wallabeng, line 531: pls use --clr-badge-info-color instead

Wallabeng (changes requested):

Wallabeng (approved):

PR #2127, opened 2024-02-29
chore: switch to Node 20, test with 18/20 (remove 16 since it is EOL)

derkoe:


package.json
```diff
@@ -102,8 +102,8 @@
     "typescript": "4.9.5"
   },
   "engines": {
-    "node": ">=16.0.0",
-    "npm": ">=7.10.0"
+    "node": ">=18.0.0",
+    "npm": ">=9.8.0"
   },
```
Wallabeng, line 107: if we allow >= 18.0.0 for node, we should allow >= 8.6.0 for npm, as this was the version bundled with that node version

Wallabeng (changes requested):

package.json
derkoe: Done

Wallabeng (approved):

PR #2176, opened 2024-04-29
if-warning Directive for clr-helper-components

gc62767:
The if-warning directive (*clrIfWarning) for the clr-helper-component should add an icon next to the form control and set the highlighting color of the helper text, as well as the control.


package.json
```diff
@@ -49,6 +49,7 @@
     "@clr/icons": "13.0.2",
     "@clr/ui": "^17.0.0",
     "@kolkov/angular-editor": "2.1.0",
+    "@ngx-translate/core": "15.0.0",
```
martinbrom, line 52: what do we need this dep for?

gc62767: Added by mistake. Will remove.

package-lock.json
```diff
@@ -22,6 +22,7 @@
         "@clr/icons": "13.0.2",
         "@clr/ui": "^17.0.0",
         "@kolkov/angular-editor": "2.1.0",
+        "@ngx-translate/core": "15.0.0",
```
Wallabeng, line 25: pls also execute "npm i" to update the lock file

src/clr-addons/clr-control-warning/if-warning.directive.ts
```diff
@@ -0,0 +1,106 @@
...
+        this.setControlStyles();
+      });
+    } else {
+      this.resetControlStyles();
+      this.icon.remove();
+      this.container.clear();
+    }
+  }
+
+  setComponents() {
```
Wallabeng, line 53: puh, thats a lot of special case handling tbh.
don't we try to add it into clarity itself to remove all that glue code which just exists, because it should belong to clarity and not addons? :)

because it needs some internal knowledge of how clarity creates its components to work correctly.

src/dev/src/app/control-warning/control-warning.demo.html
```diff
@@ -0,0 +1,171 @@
+<h1>Control Warning</h1>
```
Wallabeng, line 1: this should go into the "website" project, where the documentation resides.
the "dev" project is just a playground for developers.

Wallabeng (changes requested):

package-lock.json
gc62767: Done

Wallabeng (approved):

PR #2204, opened 2024-05-28
Website documentation for clr-control-warning

gc62767:


website/package-lock.json
```diff
@@ -53,7 +53,7 @@
     },
     "../dist/clr-addons": {
       "name": "@porscheinformatik/clr-addons",
-      "version": "17.2.1",
+      "version": "17.0.0",
```
Wallabeng, line 56: pls run "npm i" locally to update this correctly

Wallabeng (changes requested):

website/package-lock.json
Wallabeng: after building clr-addons locally

PR #2216, opened 2024-06-06
Remove max width for location bar items

jusuftopic:


src/clr-addons/location-bar/location-bar-node/location-bar-node.component.scss
```diff
@@ -5,12 +5,6 @@
       margin-bottom: 0;
     }
 
-    .dropdown {
-      .dropdown-menu {
-        max-width: 12.5rem;
```
Wallabeng, line 10: why is this removed, as it was a feature request to have it limited and shown with ellipsis

Wallabeng (changes requested):

Wallabeng (approved):

PR #2219, opened 2024-06-13
Make @kolkov/angular-editor optional

### LINKED ISSUE #1976: @kolkov/angular-editor not found
## Describe the bug

If I upgrade to version `15.3.0` or higher, I get the build error that `@kolkov/angular-editor` is not found.

## To Reproduce

Don't have `@kolkov/angular-editor` as dependency in your project and increase version to `15.3.0` or higher.

## Expected behavior

No build error.

## Versions

**App**

- Angular: 15.2.9
- Node: 18.8.0
- Clarity: 15.6.0
- Clarity-Core: 6.4.4 _(missing in template)_
- Clarity-Addons: 15.2.2 -> 15.5.6

martinbrom:
instead of providing a custom component, provide just the PHS styles

resolves #1976


martinbrom:
@Wallabeng we should properly document this in the release notes, because it's a breaking change - the old component is removed and people need to migrate to the actual angular-editor provided by the dependency (and add a CSS class).

src/clr-addons/html-editor/html/html-editor.component.ts
```diff
@@ -1,29 +0,0 @@
...
-  host: {
-    '[class.html-editor]': 'true',
-  },
-})
-export class ClrHtmlEditorComponent {
-  @Input()
-  public clrHtmlContent: string;
-  @Input()
-  public clrConfig = {
-    toolbarHiddenButtons: [['insertImage', 'insertVideo', 'fontName']],
```
Wallabeng, line 16: we should mention (at least in the patch notes) that those settings were present before.
I read your comment in the commit about this, just mentioning it here explicitly.

src/clr-addons/html-editor/html-editor.scss
```diff
@@ -0,0 +1,46 @@
+:root {
```
Wallabeng, line 1: what about adding this whole style directly to the phs theme so no further inclusion is necessary when using the phs theme? as you already use the phs colors here.
also we could apply the styles "magically" by selecting the html tag "angular-editor", so we don't have to apply a css class separately

Wallabeng (changes requested):

src/clr-addons/html-editor/html/html-editor.component.ts
martinbrom: will mention it (unless you release the version yourself)

Wallabeng (approved):

PR #2271, opened 2024-08-30
Enum filter handles undefined values

jusuftopic:


src/clr-addons/datagrid/enum-filter/enum-filter.component.ts
```diff
@@ -117,10 +144,25 @@ export class ClrEnumFilterComponent<T extends { [key: string]: any }>
...
+  private mapValues(values: (string | FilterValue)[]): FilterValue[] {
+    if (values === null || values.length === 0) {
+      return [];
+    }
+    return values.map(value => this.mapValue(value));
+  }
+
+  private mapValue(value: string | FilterValue): FilterValue {
+    if (value instanceof Object) {
+      return value.value === this.emptyDefaultValue ? { value: this.emptyValue, displayValue: this.emptyValue } : value;
```
Wallabeng, line 163: why do we need to change the "value" at all (I guess it could stay empty aka ''), shouldn't the "displayValue" be enough?
as we here set it to '(Empty)', just to reset it to '' when getting the values

Wallabeng (changes requested):

Wallabeng (approved):

src/clr-addons/datagrid/enum-filter/enum-filter.component.ts
```diff
@@ -20,34 +20,53 @@ export class ClrEnumFilterComponent<T extends { [key: string]: any }>
 {
   @Input('clrProperty') property = '';
 
+  @Input('clrEmptyValuesTranslation')
```
Wallabeng, line 23: this new input value needs some documentation on the website

Wallabeng (changes requested):

Wallabeng (approved):

PR #2286, opened 2024-09-12
fix displaying of brand avatar for brand "VW PKW"

Stronkcarry:


martinbrom:
@Stronkcarry is this still needed? I believe we discussed in Teams that you'll go with a different solution?

Stronkcarry:
not needed anymore

PR #2303, opened 2024-09-25
Angular 18 - RC 1

martinbrom:


martinbrom:
Hey guys, since Marcel is on leave for a while, can you please look this over? One open question from my side is whether we want to support Angular 17 and 18 at the same time (something like what Clarity does) or we just always keep addons-version == Angular version. I believe we've always done the latter.
Thanks

derkoe (approved):

derkoe:
> Hey guys, since Marcel is on leave for a while, can you please look this over? One open question from my side is whether we want to support Angular 17 and 18 at the same time (something like what Clarity does) or we just always keep addons-version == Angular version. I believe we've always done the latter. Thanks

Guess it's okay to only support 18.

martinbrom:
had to rebase because of Jusuf's commit that also increased all versions - jfyi will merge it without an explicit review ;)

PR #2311, opened 2024-09-30
Location bar search: Adapt min-height and set focus on open

jusuftopic:


martinbrom:
I'm assuming you need this in 17.x.x? In that case the base for your MR is wrong :)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clr-addons",
-  "version": "18.0.0-RC.1",
+  "version": "18.0.1-RC.1",
```
martinbrom, line 3: this should be v17 for you right?

martinbrom (changes requested):

PR #2322, opened 2024-10-14
Added history provider interface

hoej1:
- added interface to modify history elements in history element and pinned history

### DIFF
--- a/src/clr-addons/history/history.provider.ts
+++ b/src/clr-addons/history/history.provider.ts
@@ -4,6 +4,5 @@ import { ClrHistoryModel } from './history-model.interface';
 export const HISTORY_PROVIDER = new InjectionToken<any>('HISTORY_PROVIDER');
 
 export abstract class HistoryProvider {
-
   abstract modifyHistoryEntries(entries: ClrHistoryModel[]): void;
 }
--- a/src/clr-addons/history/history.provider.ts
+++ b/src/clr-addons/history/history.provider.ts
@@ -1,0 +1,9 @@
+import { InjectionToken } from '@angular/core';
+import { ClrHistoryModel } from './history-model.interface';
+
+export const HISTORY_PROVIDER = new InjectionToken<any>('HISTORY_PROVIDER');
+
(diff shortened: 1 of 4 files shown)


martinbrom (dismissed):

src/clr-addons/history/history.provider.ts
```diff
@@ -0,0 +1,8 @@
+import { InjectionToken } from '@angular/core';
+import { ClrHistoryModel } from './history-model.interface';
+
+export const HISTORY_PROVIDER = new InjectionToken<any>('HISTORY_PROVIDER');
+
+export abstract class HistoryProvider {
+  abstract modifyHistoryEntries(entries: ClrHistoryModel[]): void;
```
Wallabeng, line 7: pls leave a comment what this method is supposed to do. as I'm not sure what I have to expect from it :)
and I'm a big fan of return values if anything is changed, instead of modifying parameters

Wallabeng (changes requested):

src/clr-addons/history/history.provider.ts
hoej1: Done

Wallabeng:
a general question:
what is the use case for this at all? maybe we should already modify the entries before saving instead of after reading?
as with the current implementation everyone has to implement modifications after reading, to have the same behaviour on every site

and if it needs broader usage, we should add it to the documentation also

hoej1:
> a general question: what is the use case for this at all? maybe we should already modify the entries before saving instead of after reading? as with the current implementation everyone has to implement modifications after reading, to have the same behaviour on every site
> 
> and if it needs broader usage, we should add it to the documentation also

The problem here is that history entries can be written by the application in one tab and are then accessed by the application inside another tab. Depending on the use case, it might not be straight forward to share these entries between the instances running in separate tabs. So it is not sufficient to simply store the modified history entries as the modification might depend on the tab it is shown in and the history stored in the browser cookie is shared among them.

Wallabeng (approved):

PR #2323, opened 2024-10-14
Location bar feedback improvements

jusuftopic:


martinbrom:
As you already increased the version yourself, I will wait with your commit after #2322 is merged, so I can release both in one version

martinbrom (changes requested):
As you already increased the version yourself, I will wait with your commit after #2322 is merged, so I can release both in one version - this is just to block others from merging this PR, you don't need to change anything

martinbrom (approved):

PR #2330, opened 2024-10-21
Fix datagrid sticky row transparent background

martinbrom:
Because of the dg row highlighting the background-color was set to `inherit`. But that leads to seeing the row content through the sticky row if there is no highlighting applied. This fix should reset the background-color only for the highlighted rows.
PS: REQ is not even ordered yet, will release this when needed.

before:
image

after:
image


Wallabeng (approved):

d-m-s:
@martinbrom thanks for the quick fix!

martinbrom:
@d-m-s no problem :)

PR #2337, opened 2024-10-23
fix for content header

tschambalamba:


martinbrom (changes requested):
you're missing the website package-lock for version increase

martinbrom (approved):

PR #2340, opened 2024-10-24
set header min height

tschambalamba:


src/clr-addons/package-lock.json
```diff
@@ -0,0 +1,246 @@
+{
```
martinbrom, line 1: pls remove this file 😄 looks committed by mistake

martinbrom (approved):

PR #2399, opened 2024-12-04
Readonly Directive for Comboboxes, Datagrids and Select

himbeersaftmitcola:
At C3SE, we developed a readonly directive to address a common styling issue with form elements, particularly when we needed to display values in a non-editable state. While simply disabling the fields wasn't a visually appealing solution, this directive offers a much cleaner and more intuitive alternative.

This directive is now available to others as a global solution, making it easier to manage and display forms consistently.

What the Readonly Directive Does:

- Appearance: The field looks like regular text, making it clear that the value cannot be changed. Specifically, it:
- Displays the standard text color
- Has no border or outline, giving it a "read-only" appearance
- Prevents interaction, meaning the item can't be selected via mouse-click or keyboard tabbing
- Ensures the cursor remains normal when hovering over the field, indicating no editable state
- Aligns the text with other fields (those with borders) by changing padding to 0
- Behavior: The value in the field cannot be altered by the user, providing a more user-friendly and visually consistent approach than simply disabling form elements.

Visuals:
image
(See demo page)

Further information:
Combobox:
There is no implementation for multiselect yet.

Datalist:
If no value is entered and the placeholder would apply, the directive ensures that nothing appears in the field (even no placeholder) to avoid confusion, which enhances the user experience.


src/clr-addons/readonly-combobox/readonly-combobox.directive.ts
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2018-2024 Porsche Informatik. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Directive, ElementRef, Input, Renderer2 } from '@angular/core';
+
+@Directive({
+  selector: 'clr-combobox-container[readOnly], clr-datalist-container[readOnly], clr-select-container[readOnly]',
```
martinbrom, line 10: Since this is capitalized as `readOnly` I assume you want to make this directive opt-in, as opposed to it working just when using the HTML attribute `readonly` ?

If that's the case, please prefix the selector with "clr" -> `clrReadOnly` or maybe even `clrReadOnlyInput` or similar

src/clr-addons/readonly-combobox/readonly-combobox.directive.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  @Input() set readOnly(value: boolean) {
+    const inputElement = this.elRef.nativeElement.querySelector('input, select');
+    if (value) {
+      setTimeout(() => {
+        this.renderer.setAttribute(inputElement, 'disabled', 'true');
+        this.renderer.setAttribute(this.elRef.nativeElement, 'readonly', 'true');
+      });
+    } else {
+      this.renderer.removeAttribute(inputElement, 'disabled');
```
martinbrom, line 23: Do we want to support disabled and readonly elements separately? if the element was disabled before using this directive, this else branch would remove the "default" disabled state

If you say that is out of scope, then it's fine

src/clr-addons/readonly-combobox/readonly-combobox.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+      <input #hasAutoComplete autocomplete="name" clrInput type="text" />
+    </clr-input-container>
+  `,
+})
+class TestComponent {
+  @ViewChild('noAutoComplete') noAutocomplete: ElementRef;
+  @ViewChild('hasAutoComplete') hasAutocomplete: ElementRef;
+}
+
+describe('AutocompleteOffDirective', () => {
```
martinbrom, line 31: I'm assuming this is copy-pasted from somewhere? 😄 Can you adapt the tests too?

src/dev/src/app/readonly/readonly.demo.ts
```diff
@@ -14,4 +14,15 @@ export class ReadonlyDemo {
...
+  myState = {name: 'California'};
+  states = [{name: 'California'}, {name: 'Texas'}, {name: 'New York'}]
+  value = '';
+  selectedOption: any;
+
+  toggleReadonly() {
+    this.readonly = !this.readonly;
+  }
+
+  protected readonly Object = Object;
```
martinbrom, line 27: looks unused?

martinbrom (changes requested):
Do I understand it correctly, that in your use-case it's still needed for the elements to remain as form elements, for example when sending a form?. Because if not, then we could just dynamically replace the combobox/select/whatever with a literal text field.

It would save us from having to maintain the styling reset of the elements if the styling changes on Clarity side in future major releases.

src/clr-addons/readonly-combobox/readonly-combobox.directive.ts
martinbrom: Let's also rename the file and module - it's not just readonly combobox after all.

Wallabeng:
> Do I understand it correctly, that in your use-case it's still needed for the elements to remain as form elements, for example when sending a form?. Because if not, then we could just dynamically replace the combobox/select/whatever with a literal text field.
> 
> It would save us from having to maintain the styling reset of the elements if the styling changes on Clarity side in future major releases.

We also have a story in our current sprint to create a readonly directive for various UI components and our preferred way would be to replace the components with a simple span or something like that. So you would only have one "thing" to style correctly and don't need to override existing component stylings

martinbrom:
> We also have a story in our current sprint to create a readonly directive for various UI components and our preferred way would be to replace the components with a simple span or something like that. So you would only have one "thing" to style correctly and don't need to override existing component stylings

I definitely agree, but we have to hide the form component instead of replacing it, otherwise when you submit the form the fields would be missing :)

Wallabeng:
> > We also have a story in our current sprint to create a readonly directive for various UI components and our preferred way would be to replace the components with a simple span or something like that. So you would only have one "thing" to style correctly and don't need to override existing component stylings
> 
> I definitely agree, but we have to hide the form component instead of replacing it, otherwise when you submit the form the fields would be missing :)

you're right, didn't thought of this case yet (maybe when developing it)

Wallabeng:
> > We also have a story in our current sprint to create a readonly directive for various UI components and our preferred way would be to replace the components with a simple span or something like that. So you would only have one "thing" to style correctly and don't need to override existing component stylings
> 
> I definitely agree, but we have to hide the form component instead of replacing it, otherwise when you submit the form the fields would be missing :)

Is this a actually a use case? So do you submit the form directly with HTML 'action' feature?

As normally we call a service on submit in typescript code and read all the values from the angular form object. Where it shouldn't (?) be a problem when the DOM doesn't contain the element.

PR #2418, opened 2025-01-07
update readonly page style fixes and format adaptations

DaSmid:


website/src/app/documentation/demos/readonly/readonly.demo.scss
```diff
@@ -2,10 +2,18 @@
   font-weight: bold;
 }
 
+.width-100 {
```
Wallabeng, line 5: unused class, pls remove

Wallabeng (approved):

PR #2435, opened 2025-01-21
added some missed icons, some ui adaptions

tschambalamba:


website/src/styles/global.scss
```diff
@@ -307,7 +307,7 @@ a.btn-primary {
 
   .icons-content {
     position: absolute;
-    top: 15rem;
+    top: 50px;
```
d-m-s, line 310: top: 3rem;

suggest usage of rem, 3rem ~ 60px

Wallabeng (approved):

PR #2439, opened 2025-01-23
[C3DEV-317064](C3DEV-319667) Not more than 2 history entries are shown in Service

flopoi:
[C3DEV-317064](C3DEV-319667) Not more than 2 history entries are shown in Service (Es werden nicht mehr als 2 Historieneinträge im Service angezeigt) - C3REQPHSRE-4604

This reduces the Cookie length by about 33.25% on average and thus leads to the case with only 2 entries being shown happening less frequently.

Similiar Change will be contributed to cross-core-tapestry on gitlab too


flopoi:
characters outside the latin1 range cause this change to break, it won't be done anymore

PR #2457, opened 2025-02-04
add property handling for single select combobox readonly

DaSmid:


src/clr-addons/readonly/readonly.directive.ts
```diff
@@ -97,8 +97,8 @@ export class ClrReadonlyDirective implements OnChanges, OnInit, AfterViewInit {
 
     if (controlType === 'numeric') {
       return this.formatNumericValue(controlValue);
-    } else if (this.isMultiSelect) {
-      return this.formatMultiSelectValue(controlValue);
+    } else if (this.property) {
```
Wallabeng, line 100: remove the now unused input "clrMulti"

Wallabeng (changes requested):

Wallabeng (approved):

PR #2458, opened 2025-02-04
Use @use instead of deprecated @import in sass files

zsoltx:


website/src/styles/poi-specific.scss
```diff
@@ -1,3 +1,7 @@
+@use '../../node_modules/@cds/core/tokens/tokens';
```
martinbrom, line 1: If this was in the lib code I'd say to use the `var(--token)` syntax instead - but it doesn't matter as much, it's only in the website code :)

martinbrom (approved):

martinbrom:
@Wallabeng okay for you as well?

Wallabeng (approved):

Wallabeng:
> @Wallabeng okay for you as well?

lgtm!

PR #2476, opened 2025-02-19
Daterange fixes

tschambalamba:


src/clr-addons/daterangepicker/directives/daterangepicker-container/daterangepicker-container.component.ts
```diff
@@ -83,11 +79,23 @@ export class ClrDaterangepickerContainerComponent
   @Input()
   public presetsDateTime: Array<DateTimerangePreset> = [];
 
+  _timeSelection = false;
   @Input()
-  public timeSelection = false;
+  public set timeSelection(value: boolean) {
+    if (value) {
```
Wallabeng, line 85: false is also not true. i would just transform the value to a boolean and always set it
= !!value

Wallabeng (changes requested):

Wallabeng (approved):

PR #2502, opened 2025-03-17
Fix view edit section collapse animation playing on init

TobiasPankner:
Previously, the view edit section always played a expand animation in first render. 

This is very visually distracting and not desired. 

This commit disables the animation on init while keeping the animation on user interaction (collapse/expand)


TobiasPankner:
Would be nice to have in v18, but not a must-have

Wallabeng (dismissed):

Wallabeng (dismissed):

PR #2596, opened 2025-06-05
fix demo page url

tschambalamba:


martinbrom:
Why did we increase the version? 😅

Wallabeng:
because I only saw it right after committing it and then I thought I just leave it 😅

PR #2607, opened 2025-06-11
Remove margin-top from textareas to align them with labels

### LINKED ISSUE #2565: Label and textarea input are vertically misaligned
We are overriding all form controls to be aligned to "baseline" which is fine for one-line elements, but looks weird when textareas are on the same line as other inputs.

Image

martinbrom:
Resolves #2565

before:
image

after:
image
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clr-addons",
-  "version": "19.5.3",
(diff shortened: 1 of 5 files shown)


martinbrom:
fyi @d-m-s

Wallabeng (approved):

PR #2838, opened 2026-04-15
adjust copy to clipboard demo + addition of css for tooltip size

POISeunghyun:
- Adjustment of the copy to clipboard demo page, so there are more possibilities demonstrated
- As requested by Schrottshammer Dieter, addition of the clrTooltipFitContent directive for clr-tooltip-content, which sets the width of the tooltip bubble to "fit-content!important", so that the bubble is exactly as wide as the text inside.


src/clr-addons/tooltip/tooltip-fit-content.directive.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * <!-- conditionally fit-content -->
+ * <clr-tooltip-content [clrTooltipFitContent]="isFitContent" *clrIfOpen>…</clr-tooltip-content>
+ * ```
+ */
+@Directive({
+  selector: 'clr-tooltip-content[clrTooltipFitContent]',
+  standalone: true,
+})
+export class ClrTooltipFitContentDirective implements OnInit {
```
Wallabeng: can't we just provide a css class which does the same and document it?
we already did so for other stuf, it's more lightweight and doesn't need nativeElement access

Wallabeng (changes requested):

src/clr-addons/tooltip/tooltip-fit-content.directive.ts
POISeunghyun: you're right. i just took the first idea that came into my mind but a css class is definitely better. i would still keep the documentation page for tooltips though, just simplify the docu content accordingly. is that ok?

POISeunghyun:
sorry i messed up the commit history :( - its very confusing now. the only relevant commit is a9c72636594342e08a7e23cf37bf184c1e9b9dc6 & a668a05baade775608314a42c6f1493c6cdc3a0a

I hope its possible to squash the commits before merging into master?

Wallabeng (approved):

PR #2845, opened 2026-04-22
add width: fit-content

flopoi:
so the tooltip box adjusts its width according to the size of the text inside.
Currently there is a lot of space left in the box if the text is short

Current  behaviour:
grafik

After the change:
grafik


Wallabeng:
Already done in https://github.com/porscheinformatik/clarity-addons/pull/2838
Can be seen here https://porscheinformatik.github.io/clarity-addons/documentation/latest/copy-to-clipboard
With documentation of the class here https://porscheinformatik.github.io/clarity-addons/documentation/latest/tooltip

PR #2846, opened 2026-04-23
Add new image gallery component

VicSera:
This PR adds a two new components:
* `clr-image-gallery` - this is the main component that displays a static mosaic of images where one is highlighted and the rest are displayed in a thumbnail strip at the bottom. Clicking on any image will open up a modal that displays the second component
image

* `clr-image-carousel` - this is a secondary component that displays a carousel of images that can be cycled through via arrow keys, left/right buttons, or by clicking on an image in the thumbnail strip.
image

The default behavior of the `clr-image-gallery` is to spawn a built-in modal with the carousel, but optionally `[useBuiltInModal]="false"` can be specified and then `clr-image-carousel` should then be added manually. The use case for this is when the image gallery is already part of a modal and the desired approach is to replace the current modal content instead of spawning a new modal on top.


Wallabeng (dismissed):

Wallabeng (approved):

d-m-s:
Hi @VicSera, 
thanks for the great contribution! 
Please adapt the two minor issues in the vertical nav. (alphabetical sorting, remove 'new')
2026-05-19_09h28_57

PR #2854, opened 2026-05-05
[C3DEV-603168] Fixes summary area initial row state

rastidev01:
**Problem:**
The input value for the number of rows is not respected if there are not items in the summary area.
This applied to the following summary:
- data is loaded asynchonosly
- summary number of rows is set to 2
- summary loading state is connected to data loading

results in the height of the summary area starting as 3 rows and jumping to 2 rows when the data is loaded.
This is because the `currentRows` field is initialized with the default value of the `rows` input and only updated if there are item child elements in the summary area component.

**Fix:**
- Fixed by refactoring the `currentRows` field to a `linkedSignal` and connecting it to the `rows` input of the component.

### DIFF
--- a/src/clr-addons/summary-area/summary-area/summary-area.ts
+++ b/src/clr-addons/summary-area/summary-area/summary-area.ts
@@ -286,7 +286,7 @@ export class ClrSummaryArea implements OnInit, AfterViewInit, OnDestroy {
       // If enough space, use as many columns as needed for the row constraint
       // If not enough space, reduce columns and allow more rows per column
       this.currentColumns = Math.max(1, Math.min(neededColumns, maxColumnsByWidth)) as ClrSummaryAreaColumns;
-      this.currentRows.update(() => Math.ceil(itemCount / this.currentColumns) as ClrSummaryAreaRows);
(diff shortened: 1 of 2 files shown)


src/clr-addons/summary-area/summary-area/summary-area.ts
```diff
@@ -285,7 +286,7 @@ export class ClrSummaryArea implements OnInit, AfterViewInit, OnDestroy {
       // If enough space, use as many columns as needed for the row constraint
       // If not enough space, reduce columns and allow more rows per column
       this.currentColumns = Math.max(1, Math.min(neededColumns, maxColumnsByWidth)) as ClrSummaryAreaColumns;
-      this.currentRows = Math.ceil(itemCount / this.currentColumns) as ClrSummaryAreaRows;
+      this.currentRows.update(() => Math.ceil(itemCount / this.currentColumns) as ClrSummaryAreaRows);
```
Wallabeng: just use .set(..) for static values, when you don't need the current value of the signal

Wallabeng (changes requested):

Wallabeng (approved):

PR #2901, opened 2026-06-17
Collapse/Expand: Missing "Title"

markofilipovicext:
…arity addons

- The missing title was caused by projecting [clr-ces-title] through ng-template + *ngTemplateOutlet — that pattern breaks under Angular 21’s control flow. -- Title and subtitle are now projected directly in each @if branch -- Only the caret button stays in a shared #caretBtn template (no ng-content inside it)

- A unit test was added to assert title, subtitle, and content projection.


Wallabeng (approved):

d-m-s:
@markofilipovicext, @himbeersaftmitcola
See screenshot; please make both title (1) and caret (2) clickable:
2026-06-17_14h28_04

PR #2907, opened 2026-06-17
Collapse/Expand: Title should be clickable

markofilipovicext:
…arity addons

- Make the title clickable, same behavior as the chevron.

### DIFF
--- a/src/clr-addons/collapse-expand-section/collapse-expand-section.html
+++ b/src/clr-addons/collapse-expand-section/collapse-expand-section.html
@@ -2,11 +2,31 @@
   <div class="ces-title" [class.disabled-header-styles]="disableHeaderStyles">
     @if (!disableHeaderStyles) {
     <h2>
-      <ng-content select="[clr-ces-title]"></ng-content>
+      <span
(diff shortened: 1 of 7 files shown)


src/clr-addons/collapse-expand-section/collapse-expand-section.html
```diff
@@ -2,11 +2,25 @@
   <div class="ces-title" [class.disabled-header-styles]="disableHeaderStyles">
     @if (!disableHeaderStyles) {
     <h2>
-      <ng-content select="[clr-ces-title]"></ng-content>
+      <button
```
Wallabeng: isn't a clickable div or span easier than a button, where you have to remove all styles of the button

Wallabeng (changes requested):

src/clr-addons/collapse-expand-section/collapse-expand-section.html
markofilipovicext: Updated.

Wallabeng (approved):

PR #2926, opened 2026-06-30
C3DEV-648296 fix bug in collapse expand section

POISeunghyun:
collapse expand section does not show any title if clrDisableHeaderStyles is true
this fix makes the component behave as it is supposed to, when clrDisableHeaderStyles is true

PR #2937, opened 2026-07-07
Fix collapse expand section clrDisableHeaderStyles

TobiasPankner:
Currently no title or subtitle is shown when clrDisableHeaderStyles=true

- Broken with the control flow migration
- This did not work anymore as only the first "select" is filled by Angular


Wallabeng:
closed as duplicate to https://github.com/porscheinformatik/clarity-addons/pull/2926

PR #2938, opened 2026-07-07
[C3DEV-575793](C3DEV-611869) Adds hideable columns to treetable

rastidev01:
- Adds Clairty Datagrid like hideable columns to the treetable component and feature necessary services, directives, components, ... etc.
- Fixes font-size of treetable to be the same as datagrid (14px instead of 13px).
- Refactors main-renderer to use signal based functionality and support column visibility.
- Refactors some service names to follow naming of other services in the treetable. 
- Updates documentation.


src/clr-addons/treetable/providers/treetable-column-state.service.ts
```diff
@@ -0,0 +1,229 @@
...
+export const TreetableColumnUpdate = {
+  WIDTH: 'WIDTH',
+  HIDDEN: 'HIDDEN',
+  RESET_HIDDEN: 'RESET_HIDDEN',
+} as const;
+export type TreetableColumnUpdate = (typeof TreetableColumnUpdate)[keyof typeof TreetableColumnUpdate];
+
+@Injectable()
+export class TreetableColumnStateService {
+  private readonly _changeWidth$ = new Subject<{ id: string; width: number }>();
```
Wallabeng, line 24: whats the intention behind the mixture of rxjs and signals?
wouldn't it be more readable with signals only?

rastidev01: This is needed because there are async changes that need to happen at the same time. Because the changes$ Observable needs to emit the type of change, so that the renderer can use the appropriate rendering rules.

We could do this with signals as well, but I don't think it would be a lot more readable/easier. We would need a signal for the changes, wire the _columnState with a linkedSignal and merge toObservables of the signals into the current changes$ Observable.

Wallabeng (approved):

PR #2955, opened 2026-07-14
Fix with of input type time

DaSmid:


src/clr-addons/components.clr-addons.scss
```diff
@@ -409,6 +409,11 @@ clr-date-container .clr-control-container {
   }
 }
 
+//Set the width of all inputs with the class time-width
+.clr-input-group:has(.time-width) {
```
Wallabeng: time-width is only used in documentation. it's not a css class of clarity or clarity-addons library

Wallabeng (changes requested):

src/clr-addons/components.clr-addons.scss
DaSmid: change to type

src/clr-addons/components.clr-addons.scss
```diff
@@ -409,6 +409,11 @@ clr-date-container .clr-control-container {
   }
 }
 
+//Set the width of all inputs with the class time-width
```
Wallabeng: pls adapt comment

Wallabeng (changes requested):

Wallabeng (approved):

PR #2959, opened 2026-07-16
Add badge to card title of view-edit-section

alexteo24:
This PR added the option to add a badge next to title in `clr-view-edit-section`

The badge could be added in multiple ways, provided that `clrShowBadge` is set to true

- via `clrBadgeRef`, which allows for projection of content via ng-template
image

- via `clrBadgeContent` and `clrBadgeClass`. If `clrBadgeClass` is not specified, the styles of label and label-primary will be used
image

### DIFF
--- a/src/clr-addons/view-edit-section/view-edit-section.html
+++ b/src/clr-addons/view-edit-section/view-edit-section.html
@@ -1,6 +1,13 @@
 <div class="card view-edit-section" (keyup.escape)="onCancel()">
   <div class="card-block">
     <div class="card-title" [attr.cds-text]="_titleTextStyle + ' ' + _titleTextFontWeight">
+      <ng-template #titleContainer>
+        {{ _title }} @if (_showBadge) { @if (_badgeContent) {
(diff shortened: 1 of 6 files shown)


src/clr-addons/view-edit-section/view-edit-section.html
```diff
@@ -1,6 +1,17 @@
 <div class="card view-edit-section" (keyup.escape)="onCancel()">
   <div class="card-block">
     <div class="card-title" [attr.cds-text]="_titleTextStyle + ' ' + _titleTextFontWeight">
+      <ng-template #titleContainer>
+        {{ _title }} @if (_showBadge) { @if (_badgeRef) {
+        <span class="ves-badge">
+          <ng-template [ngTemplateOutlet]="_badgeRef"></ng-template>
```
Wallabeng: as you already have content projection below, the badgeRef input is redundant and brings no further use, right?

Wallabeng (changes requested):

Wallabeng (approved):

PR #2969, opened 2026-07-21
C3DEV-659339 fix bug: wrong tenant application dropdown height with clr/angular v18+

POISeunghyun:
In Cross3, the tenan application dropdown of the menu bar has been observed to be very short. upon analysis it turned out to be because of the upgrade from clr/angular v17 to v18+
- @clr/angular v18+ switched from Renderer2 (synchronous) to CDK Overlay (async requestAnimationFrame), so positioning is no longer guaranteed to be complete by the time ngAfterViewChecked runs in ClrDropdownOverflowDirective.
- this change will not break the existing behaviour for applications using clr/angular v17 (e.g. service-base currently)

### DIFF
--- a/src/clr-addons/dropdown/clr-dropdown-overflow.directive.ts
+++ b/src/clr-addons/dropdown/clr-dropdown-overflow.directive.ts
@@ -4,7 +4,16 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { AfterViewChecked, ContentChildren, Directive, ElementRef, Input, OnDestroy, QueryList } from '@angular/core';
+import {
+  AfterViewChecked,
+  ContentChildren,
+  Directive,
+  ElementRef,
+  Input,
+  NgZone,
+  OnDestroy,
+  QueryList,
+} from '@angular/core';
 import { ClrDropdown } from '@clr/angular';
 import { Subscription } from 'rxjs';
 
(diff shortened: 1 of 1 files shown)


src/clr-addons/dropdown/clr-dropdown-overflow.directive.ts
```diff
@@ -59,13 +69,27 @@ export class ClrDropdownOverflowDirective implements AfterViewChecked, OnDestroy
...
+    // If the element is at position 0,0 it may not yet be positioned by the popover/CDK overlay system.
+    // Note: @clr/angular v18+ switched from Renderer2 (synchronous) to CDK Overlay (async rAF),
+    // so positioning is no longer guaranteed to be complete by the time ngAfterViewChecked runs.
+    // Defer the check to allow async positioning (e.g. CDK Overlay) to complete.
+    if (rect.x === 0 && rect.y === 0 && !this.alreadyChecked && !this.deferredCheckPending) {
+      this.deferredCheckPending = true;
+      this.ngZone.runOutsideAngular(() => {
+        requestAnimationFrame(() => {
+          this.deferredCheckPending = false;
+          this.ngZone.run(() => this.applyDropdownOverflowStyles());
```
Wallabeng, line 92: do we need some potential endless loop exit like trying 10 times and then give up?

Wallabeng (changes requested):

src/clr-addons/dropdown/clr-dropdown-overflow.directive.ts
POISeunghyun: good point. i assume the position x=0 and y=0 on a dropdown is rare but not impossible. i will add a retry counter and exit the loop after 10 tries

POISeunghyun:
@Wallabeng can i add the prettier check fix here? the formatting errors are from a lot of files i didnt modify, so i want to make sure its ok to push them here

Wallabeng:
> @Wallabeng can i add the prettier check fix here? the formatting errors are from a lot of files i didnt modify, so i want to make sure its ok to push them here

sure go for it

POISeunghyun:
seems like i had a wrong version for the formatting libraries on my local environment 😄 build passed now!

Wallabeng (approved):

PR #2982, opened 2026-07-28
add grouped bar chart

VK93YQ9:
added grouped bar charts that can display horizontal or vertical

### DIFF
--- a/src/dev/src/app/app.routing.ts
+++ b/src/dev/src/app/app.routing.ts
@@ -14,6 +14,10 @@ export const APP_ROUTES: Routes = [
     loadComponent: () => import('./bar-chart/bar-chart.demo').then(m => m.BarChartDemo),
   },
   {
+    path: 'grouped-bar-chart',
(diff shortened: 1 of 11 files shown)


src/dev/src/app/app.routing.ts
```diff
@@ -13,6 +13,10 @@ export const APP_ROUTES: Routes = [
     path: 'bar-chart',
     loadComponent: () => import('./bar-chart/bar-chart.demo').then(m => m.BarChartDemo),
   },
+  {
+    path: 'grouped-bar-chart',
+    loadComponent: () => import('./grouped-bar-chart/grouped-bar-chart.demo').then(m => m.GroupedBarChartDemo),
```
Wallabeng, line 18: seems you forgot that code to commit

Wallabeng (changes requested):

VK93YQ9:
sorry for the late response, i just came back from vacation. Yes indeed there wer some unversioned files that i forgot to commit.

Wallabeng (approved):

PR #3017, opened 2026-08-18
Fix: Filter out unordered columns in tables when the order attribute does not exist

Flavius10:
Fix: Filter out unordered columns in tables when the order attribute does not exist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clr-addons",
-  "version": "21.5.3",
+  "version": "21.5.4",
   "private": true,
   "scripts": {
     "build:ui:css": "sass --source-map --precision=8 ./src/clr-addons/themes/phs/phs-theme.scss ./dist/clr-addons/styles/clr-addons-phs.css",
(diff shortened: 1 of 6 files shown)


src/clr-addons/datagrid/datagrid-state-persistence/state-persistence-key.spec.ts
```diff
@@ -30,18 +30,16 @@ import { ClrDatagridStatePersistenceModel } from './datagrid-state-persistence-m
...
       [clrDatagridColumnReorder]="columns"
       (clrDatagridColumnOrderChanged)="columns = $event.columns"
       (clrDgRefresh)="refreshHandler.next($event)"
     >
-      @for (col of columns; track col.name) {
-        <clr-dg-column [clrDgField]="col.name" cdkDrag>
-          <ng-container *clrDgHideableColumn="{ hidden: col.hidden }">{{ col.title }}</ng-container>
-        </clr-dg-column>
-      }
+      <clr-dg-column *ngFor="let col of columns" [clrDgField]="col.name" cdkDrag>
```
Wallabeng: pls keep the @for form

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clr-addons",
-  "version": "21.5.3",
+  "version": "21.5.4",
```
Wallabeng, line 3: if you update the version, also package-lock.json and website/package.json must be updated.
- npm i
- npm run build
- cd website
- npm i

Wallabeng (changes requested):

Wallabeng (approved):

PR #3021, opened 2026-08-19
Fix: Filter out unordered columns in tables when the order attribute does not exist

Flavius10:
Fix: Filter out unordered columns in tables when the order attribute does not exist


Wallabeng (changes requested):
I don't think this was intended like that :-)

PR #3057, opened 2026-09-08
Set min-width on sortable/filterable datagrid columns & cells

s2philarb:
Enforce a minimum width on datagrid columns that contain a sort icon or filter button so their headers don't get squeezed.

Before:
image

After:
image


Wallabeng (changes requested):
As discussed we should discuss this :D

PR #3071, opened 2026-09-14
fix: suppress Clarity18 hover bar on non-active nav items

s2philarb:
After upgrading to Clarity 18, hovering over navigation and menu items makes them look **selected**. A hovered item becomes visually the same as the actually active item.

- Header menu bar: hovering "Configuration" looks the same as the active
  "Systemmanagement" item
- Dropdown filter popovers (e.g. the funnel filter in Communicator): the
  trash ("Clear filters") and close icons show a green left-edge bar on hover

## Cause

Clarity 18 added a left-edge highlight indicator rendered via a `::before` (an inset `border-left`) on interactive nav/menu items.

In order to maneuver around this beahvior I added a global override in the phs theme that keeps the highlight bar for the active item but makes it transparent on hover of non-active items. Dropdown action buttons have no active state, so their hover indicator is simply removed.


src/clr-addons/themes/phs/_css.overrides.scss
```diff
@@ -312,6 +312,16 @@
...
   }
+
+  // Clarity 18 draws a left-edge highlight bar (a ::before with border-left) on nav/menu items for both
+  // :hover and .active using the same color. Keep the bar for the active item but suppress it on hover of non-active
+  // & items that should not get this effect in the first place within the dropdown
+  .nav-link:hover:not(.active)::before,
+  .clr-vertical-nav .nav-group-content:hover::before,
+  .dropdown-menu .dropdown-item:hover:not(.active)::before,
+  .dropdown-menu .btn:hover::before {
+    border-color: transparent !important;
```
Wallabeng: can't we override the used css3 variable used in clarity for these elements?

Wallabeng (changes requested):

Wallabeng (approved):

